Issue #223, opened 2014-03-31
Add support to .phpbrewrc


hlegius:
Hi folks,

Would be nice if phpbrew has a `.phpbrewrc` (rvm `.rvmrc` clone) or something to allow us configure what PHP's version should be loaded when entering in some project directory.

brunobatista:
:+1:

marcioAlmada:
@brunobatista say thanks to @morozov https://github.com/phpbrew/phpbrew/pull/391;)

PR #224, opened 2014-04-12
W.I.P support for .phpbrewrc #223

marcioAlmada:
Maybe:
- [ ] a way to restore php version after leaving dir with .phpbrewrc
- [ ] general hooks to be used inside .phpbrewrc files (ondir, onleave)


brunobatista:
+1

PR #391, opened 2014-11-11
Support for .phpbrewrc

morozov:
This is a proof-of-concept which works with Bash on Ubuntu 14.10 x64. I didn't test any other configurations.

The `echo` statements may be useful for testing, but I believe they are worth removing after the implementation is approved.

The original idea was borrowed from https://github.com/phpbrew/phpbrew/pull/233.


brunobatista:
Thanks a lot @morozov :+1:

PR #444, opened 2015-01-01
Fixes a bug that donwloading a file using wget or curl always fails even when it succeeds.

shinnya:
## Purpose
To resolve #443 .

## Changes
Use `Utils::system` method to download a file when wget or curl is used in `UrlDownloader`.

## Effect
Users who don't enable curl extension can download files correctly.

## Tests
Add some test cases for `UrlDownloaderTest`.

PR #445, opened 2015-01-01
Removes redundant if statements in UrlDownloader.

shinnya:
## Purpose
To remove redundant if statements which should be removed in #444 . `Utils::system` method throws an exception if command fails, so there is no need to throw an exception after calling `Utils::system`.


shinnya:
I know I should send PR to develop branch, but develop branch doesn't merge the previous PR yet.

c9s:
looks good! thanks!

Issue #464, opened 2015-02-03
"name lookup timed out" error


vasiliy-pdk:
Hi! I get "name lookup timed out" message when I try to fetch known versions or to install some php.

Here is the output: 

```
vasiliy@vasiliy-Inspiron-7720:~$ phpbrew -v -d known
===> Fetching release list...
name lookup timed out
vasiliy@vasiliy-Inspiron-7720:~$ phpbrew -v -d install 5.6.5 +default
name lookup timed out
vasiliy@vasiliy-Inspiron-7720:~$ phpbrew --version
phpbrew - 1.18.4
cliframework core: 1.10.0
```

You can see my php.ini here https://www.dropbox.com/s/0xlv6y2cts46u08/php.ini?dl=0
You can see output of the phpinfo here: https://www.dropbox.com/s/x56kcnp9c0w6s7s/phpinfodump?dl=0

The problem appears on my Ubuntu 12.04. I have tried to disable Suhosin plugin by 

```
[suhosin]
suhosin.simulation = On
```

c9s:
looks like your network is blocked? try using proxy options?

vasiliy-pdk:
I am going to try proxy. Where can I find a URL requested by phpbrew's commands `<known>` or `<install>`?

vasiliy-pdk:
Network is not blocked. I've found the URL "https://raw.githubusercontent.com/phpbrew/phpbrew/$branch/assets/php-releases.json" here https://github.com/phpbrew/phpbrew/blob/develop/src/PhpBrew/ReleaseList.php#L103 .

I've copied it to the browser's address bar and replaced `$branch` with "master". I've seen correct packages json structure. So, I don't think that the problem appears because of network.

vasiliy-pdk:
I've found the problem. The reason for the error is the small value of CURLOPT_CONNECTTIMEOUT. It is set to 10 seconds now. When I set it to 30 seconds, all works fine.  

There is no way to customize it by any option of PHPBrew now. 

Here is set the option: https://github.com/c9s/CurlKit/blob/1.0.5/src/CurlKit/CurlDownloader.php#L72
`CurlDownloader` allows us to set an another value. But we don't do this in the `ReleaseList::fetchRemoteReleaseList` method: https://github.com/phpbrew/phpbrew/blob/develop/src/PhpBrew/ReleaseList.php#L106

So, I suggest to add `curl-connection-timeout` option to the PHPBrew's Config file. 

[...]

PR #465, opened 2015-02-03
'--downloader-connect-timeout' option has been added to several commands

vasiliy-pdk:
`--downloader-connect-timeout` option has been added to these commands `known`, `update`, `install`, `download`.

## Purpose

To fix "name lookup timed out" error #464

## Changes

See changed files below.

## Effect

This option with description will be showed with a help message.

## Tests

Test cases related to this PR had been failing before I made any changes to the code. So I did not write tests.
I need more information about deployment of the development version to run tests.

## Usage

If you get "name lookup timed out" error during execution of any command from the set: `known`, `update`, `install`, `download`. You can set additional option --downloader-connect-timeout=seconds to fix the error.

For example `--downloader-connect-timeout=30`

### DIFF
--- a/src/PhpBrew/Command/DownloadCommand.php
+++ b/src/PhpBrew/Command/DownloadCommand.php
@@ -34,8 +34,9 @@ class DownloadCommand extends Command
         $opts->add('old', 'enable old phps (less than 5.3)');
         $opts->add('mirror:', 'Use mirror specific site.');
         
-        $opts->add('downloader-connect-timeout:', 'The number of seconds for '
-                . 'CURLOPT_CONNECTTIMEOUT option')
+        $opts->add('connect-timeout:', 'The system aborts the command if downloading '
+                . 'of a php version not starts during this limit. This option '
+                . 'overrides a value of CONNECT_TIMEOUT environment variable.')
             ->valueName('seconds')
             ;
     }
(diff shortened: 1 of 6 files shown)


c9s:
Can we rename the option `--downloader-connect-timeout=30` into `--curl-connect-timeout=30` and also add read the `CURL_CONNECTTIMEOUT` environment variable to override the default setting ?

vasiliy-pdk:
Actually I've named the option `--curl-connect-timeout` for the first time :) But then I've renamed it because I thought that `downloader` will be more appropriate abstract name for a property of `Downloader` abstract class. `CurlDownloader` may be a particular subclass extending the `Downloader`. 

If you think that we should call the option `--curl-connect-timeout` I can rename it. 

As for environment variable I am going to think about it today or tomorrow.

vasiliy-pdk:
I've added ability to set a value of connect-timeout by `DOWNLOADER_CONNECT_TIMEOUT` environment variable. This value will be overridden by a value of `--downloader-connect-timeout` option if it providing. 

Let me know if you want to rename option and env variable please.

c9s:
Sounds reasonable, can we just remove the `downloader` ? so the option name could be shorter: `--connect-timeout=60`.

vasiliy-pdk:
I've made another PR #466 . It contains the changes from current PR and a fix for the fatal error during tests running. So there are passed tests.

PR #466, opened 2015-02-04
CONNECT_TIMEOUT env variable and --connect_timeout option was added. Fatal Error during tests fixed

vasiliy-pdk:
There was an error with the tests execution. 

```
PHP Fatal error:  Class PhpBrew\Tasks\MakeTaskTestBuild contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (PhpBrew\Buildable::isBuildable) in /home/travis/build/phpbrew/phpbrew/tests/PhpBrew/Tasks/MakeTaskTest.php on line 70
```

I've fixed it.


c9s:
Thanks!

PR #498, opened 2015-05-12
Saving some memory for echo.

dseguy:


shinnya:
Thank you for your contribution, but it doesn't make sence for me to save such a small memory consumption, so could you describe why you think it meaningful?

dseguy:
marcio : tell me if you prefer this PR on another branch.

dseguy:
@shinnya : It's a simple change, it takes advantage of PHP features and you don't have to think about it. 
I agree it's a small step, though it is still a step.

shinnya:
I can't understand why I don't have to think about the reason, but the following code indicates that there is no difference in terms of the memory comsumption.

```php
<?php

$before = memory_get_usage();

// use comma => 184 in my environment
echo 'a', "\n";
echo "memory usage: ", memory_get_usage() - $before, "\n";

// use . => 184 in my environment
echo 'b' . "\n";
echo "memory usage: ", memory_get_usage() - $before, "\n";
```

c9s:
Hi @dseguy , thanks for your work. using comma does boost the output performance for micro-optimization (when N > 10M) , however we dont have so many message to output here. we'd like to receive PR that's planned or discussed, hope you dont mind. :) Thanks for your work.

c9s:
also, PRs should be rebased from master to avoid confliction.

git pull --rebase origin master

the above command will help you do the work.

Issue #487, opened 2015-04-20
phpbrew prepends $PATH on every cd command


larsnystrom:
I believe PR #391 introduced a bug which causes `$PATH` to be prepended with the path to the current php version every time you run `cd`. Example output from my terminal:

	lars@lars-zb-ubuntu:~$ echo $PATH
	/home/lars/.phpbrew/php/php-5.5.24/bin:/home/lars/.phpbrew/php/php-5.5.24/bin:/usr/local/heroku/bin:/home/lars/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/lib/go/bin:/home/lars/.go/bin
	lars@lars-zb-ubuntu:~$ cd ..
	lars@lars-zb-ubuntu:/home$ echo $PATH
	/home/lars/.phpbrew/php/php-5.5.24/bin:/home/lars/.phpbrew/php/php-5.5.24/bin:/home/lars/.phpbrew/php/php-5.5.24/bin:/usr/local/heroku/bin:/home/lars/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/lib/go/bin:/home/lars/.go/bin
	lars@lars-zb-ubuntu:/home$ cd
	lars@lars-zb-ubuntu:~$ echo $PATH
	/home/lars/.phpbrew/php/php-5.5.24/bin:/home/lars/.phpbrew/php/php-5.5.24/bin:/home/lars/.phpbrew/php/php-5.5.24/bin:/home/lars/.phpbrew/php/php-5.5.24/bin:/usr/local/heroku/bin:/home/lars/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/lib/go/bin:/home/lars/.go/bin

The path is also prepended twice when you start a new shell.

We should use `sed` or something else to replace the current path instead of just prepending the path all the time.

marcioAlmada:
confirmed cc @morozov

morozov:
I'll take a look at this probably next week.

morozov:
[...]

PR #501, opened 2015-05-14
export $PHPBREW_LAST_DIR at the bottom of _phpbrewrc_load.

shinnya:
CC @morozov

I think https://github.com/phpbrew/phpbrew/pull/391 has the issue that _phpbrewrc_load is executed on every shell command because the function doesn't export `PHPBREW_LAST_DIR`. https://github.com/phpbrew/phpbrew/issues/487 might be related to this behaviour(?)
To solve this issue, we should export `$PHPBREW_LAST_DIR` after _phpbrewrc_load to prepare for the next call.

### DIFF
--- a/shell/bashrc
+++ b/shell/bashrc
@@ -494,7 +494,7 @@ function _phpbrewrc_load ()
         OLDPWD="$prev_dir"
     fi
 
-    PHPBREW_LAST_DIR="$PWD"
+    export PHPBREW_LAST_DIR="$PWD"
 }
 
 if [[ -n "$BASH_VERSION" && -z "$PHPBREW_RC_DISABLE" ]]; then


shinnya:
I guess the reporter of https://github.com/phpbrew/phpbrew/issues/487 would re-source `.phpbrew/bashrc` on every cd command.

morozov:
@shinnya is exporting `$PHPBREW_LAST_DIR` supposed to fix the issue when `.phpbrew/bashrc` is re-sourced on every `cd` command? Why would one do that in the first place?

I admin, exported `$PHPBREW_LAST_DIR` may be useful in some cases, but in any case, using some smart function like `__phpbrew_set_path()` which does its job (if that's is job) is still better than ` PATH=$PHPBREW_PATH:$PATH` w/o taking into account what the current value of `$PATH` is.

shinnya:
@morozov I thought `_phpbrewrc` would pass through the condition of https://github.com/phpbrew/phpbrew/blob/issue/487/shell/bashrc#L458 without exporting `$PHPBREW_LAST_DIR`, but I'm sorry it's my misunderstanding. Then, I didn't have intention to use `$PHPBREW_LAST_DIR` for other purposes like what you explained.

Issue #506, opened 2015-05-20
command too long: phpbrew()


greensn0w:
I just installed phpbrew on OSX (latest).
I'm using ZSH with zprezto.

When running `phpbrew` i get this output:

[...]

PR #508, opened 2015-05-20
Add "-p" to $(command which phpbrew) to ensure a default PATH

greensn0w:
The `-p' option means to use a default value for $PATH that is guaranteed to find all of the standard utilities
This also fixes the "env: node: No such file or directory" issue I mentioned at #506 (last comment)

### DIFF
--- a/shell/bashrc
+++ b/shell/bashrc
@@ -97,7 +97,7 @@ function phpbrew ()
     if [[ -e bin/phpbrew ]] ; then
         BIN='bin/phpbrew'
     else
-        BIN=$(command which phpbrew)
+        BIN=$(command -p which phpbrew)
     fi
 
     local exit_status


c9s:
Can you make sure that `-p` works for both bash/zsh on OS X and Linux? Thanks

c9s:
ah, I tested, it works. Thanks!

marcioAlmada:
This fails when phpbrew is at `/usr/local/bin`. Can we revert?

marcioAlmada:
```bash
$ command -p which phpbrew
which: no phpbrew in (/bin:/usr/bin)
```

marcioAlmada:
This command short circuit works form me:

`BIN=$(command -p which phpbrew 2> /dev/null || command which phpbrew)`

PR #511, opened 2015-05-21
Extend $PATH instead of overwriting it

greensn0w:
Before this change phpbrew clears your path and you can do nothing until you remove phpbrew's bashrc


marcioAlmada:
thanks!

PR #514, opened 2015-05-27
Release PHP 5.6.9

ovr:


marcioAlmada:
Please don't update this release list file anymore, we don't need it. Just `phpbrew self-update` and then run `phpbrew update` so phpbrew can get the release list directly from the php.net website.

Thanks for the pull request, anyway.

Issue #520, opened 2015-06-08
Installing extensions from github


uda:
I was trying to install an extension from github. but apparently phpbrew uses the old path: `https://github.com/someuser/somerepo/tarball/branch`, which yields a 404 page.

Also, using the URL with `.git` extension also yields an error, so the documentation should be updated.

marcioAlmada:
Did the url changed recently? this worked a few days ago. I'll check your pull request :)

marcioAlmada:
FYI `.git` is working too.

PR #521, opened 2015-06-08
Get correct Github repository tarballs and extracted directory

uda:
This fixes Issue #520


marcioAlmada:
Thanks for noticing that. Release will come ASA travis confirms all green.

PR #531, opened 2015-06-22
update doc for known command, add --http-proxy and --http-proxy-auth

### LINKED ISSUE #530: HTTP proxy in known command
Option --http-proxy is available for `install` command but not for `known` and `update`. Maybe even more commands should support it (global option?).

jhdxr:
fix #530


jhdxr:
the implements of `--http-proxy` and `--http-proxy-auth` already exist in `ReleaseList::buildReleaseListFromOfficialSite`, so I just update the doc.

marcioAlmada:
hey, thank you, we didn't notice that.

PR #569, opened 2015-08-27
Fix curl error handling when self updating

konomae:
Sometimes, curl fails some reason.
like this

```bash
$ phpbrew self-update

curl: (60) SSL certificate problem: Invalid certificate chain
More details here: http://curl.haxx.se/docs/sslcerts.html
```

But the fail ignored, and overwrite `phpbrew` binary with failed file.

### DIFF
--- a/src/PhpBrew/Command/SelfUpdateCommand.php
+++ b/src/PhpBrew/Command/SelfUpdateCommand.php
@@ -41,8 +41,8 @@ class SelfUpdateCommand extends Command
             throw new RuntimeException("Fail to create temp file", 2);
         }
         chmod($tempFile, 0755);
-        system("curl -# -L $url > $tempFile", $code);
-        if(! $code == 0) {
+        $lastLine = system("curl -# -L $url > $tempFile", $code);
+        if($lastLine === false || ! $code == 0) {
             throw new RuntimeException("Update Failed", 1);
         }
         //todo we can check the hash here in order to make sure we have download the phar successfully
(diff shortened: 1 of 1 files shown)


jhdxr:
the manual says the `system` function 
> Returns the last line of the command output on success, and FALSE on failure. 
so it's why the origin script fails on `$code == 0`

However, you modified the script and set `$code` to `return_var`, and keep the sentences which throw an exception if $code equals 0. `return_var` equals 0 should means everything is ok, and NO exception should be thrown.

konomae:
> return_var equals 0 should means everything is ok, and NO exception should be thrown.

`if(! $code == 0)` has `!`.
If `return_var` is 0, `!(0 == 0) === false`. No exception thrown.

And `system` function returns empty string if curl fails.

```php
<?php
$tempFile = tempnam(sys_get_temp_dir(), 'phpbrew_');
$code = system("curl -# -L https://raw.githubusercontent_.com/phpbrew/phpbrew/master/phpbrew > $tempFile", $retValue);
var_dump($code, $retValue);

if(! $code == 0) { // false
	throw new RuntimeException("Update Failed 1", 1);
}

if(! $retValue == 0) { // true
	throw new RuntimeException("Update Failed 2", 1);
}
```
Output:

```
curl: (6) Could not resolve host: raw.githubusercontent_.com
string(0) ""
int(6)
PHP Fatal error:  Uncaught exception 'RuntimeException' with message 'Update Failed 2' in /path/to/test.php:11
Stack trace:
#0 {main}
  thrown in /path/to/test.php on line 11

Fatal error: Uncaught exception 'RuntimeException' with message 'Update Failed 2' in /path/to/test.php on line 11
```

jhdxr:
orz I ignored `!`...

konomae:
Added check for return value of `system` function

PR #585, opened 2015-09-28
Import of class PhpBrew\Utils in PhpBrew\Command\CleanCommand

dharmavagabond:
When using ```phpbrew clean --all php-x.y.z```, it looks for ```PhpBrew\Command\Utils``` which doesn’t exists.
```php 
Source directory /Users/Dharma/.phpbrew/build/php-5.6.12 found, deleting...
PHP Fatal error:  Class 'PhpBrew\Command\Utils' not found in phar:///Users/Dharma/bin/phpbrew/PhpBrew/Command/CleanCommand.php on line 2```


marcioAlmada:
Good catch. Ty.

PR #588, opened 2015-10-02
Fix issues with fpm

shinigamicorei7:
Sudo simply increase on some lines.
Only proven ubuntu.


c9s:
I am not sure if this is a good fix, @marcioAlmada what do you think?

PR #596, opened 2015-10-18
 Bash prompt, php version was never updated

gsouf:
The bash prompt was never updated.

Before the change:

```bash
~ > me@debian [php:5.6.13-system]
$  phpbrew use 5.4.45
~ > me@debian [php:5.6.13-phpbrew]
$ 
```

After the change:

```bash
~ > me@debian [php:5.6.13-system]
$  phpbrew use 5.4.45
~ > me@debian [php:5.4.45-phpbrew]
$ 
```
```


gsouf:
How fast...

c9s:
:D

Issue #512, opened 2015-05-26
Implement an argument that prints the build output


greensn0w:
This is more an idea than an issue.
I think we should implement an argument like `--verbose-build` that prints the configure and make.
It's really annoying to copy/paste the tail command every time you installed a missing dep.

Let's discuss this first, I'm very excited to hear what you think about this :smile: 

If we decide to include an argument like this, I can take over the part of coding this feature :smiley:

marcioAlmada:
have you tried `--stdout` option? `phpbrew install --stdout ...`

marcioAlmada:
Please reopen the issue if it's not what you want or if you have a better idea :)

greensn0w:
It's nice to have this option but there are a few issues with it:
1. It's not in `phpbrew help`
2. It's name is confusing because `--stdout` could simply mean a `2>&1` alias
3. It's not working:
```shell
greensn0w@Mac  ~ 
 ❯ phpbrew install --stdout 5.3.29 +default
===> phpbrew will now build 5.3.29
PHP 5.3 requires +intl, enabled by default.
===> Loading and resolving variants...
Checking distribution checksum...
Checksum matched: 9469e240cbe6ac865aeaec89b253dd30
===> Extracting /Users/greensn0w/.phpbrew/distfiles/php-5.3.29.tar.bz2 to /Users/greensn0w/.phpbrew/tmp/php-5.3.29
===> Removing /Users/greensn0w/.phpbrew/build/php-5.3.29
===> Moving /Users/greensn0w/.phpbrew/tmp/php-5.3.29 to /Users/greensn0w/.phpbrew/build/php-5.3.29
===> Configuring 5.3.29...


Use tail command to see what's going on:
   $ tail -F /Users/greensn0w/.phpbrew/build/php-5.3.29/build.log


tee: illegal option -- -
usage: tee [-ai] [file ...]
./configure: line 1866: echo: write error: Broken pipe
./configure: line 1895: echo: write error: Broken pipe
./configure: line 1925: echo: write error: Broken pipe
```

marcioAlmada:
> 1\. It's not in phpbrew help

It is, try `phpbrew help install`

> 2\. It's name is confusing because --stdout could simply mean a 2>&1 alias

I'd accept a pull request with --verbose then, I can't remember why it's not called `--verbose`... @c9s?

> 3\. It's not working:

Works with bash here. Can you describe your environment?

c9s:
[...]

PR #598, opened 2015-10-20
Make `tee` call compatible with OSX

augustohp:
OSX's `tee` command do not accept long option formats like `--append`,
only their short forms. They probably needed to save the memory for some
huge background image, or were just willingly to break stuff (which is
more probably).

The bug occurs on OSX when installing a new PHP version using --stdout
option.

### DIFF
--- a/src/PhpBrew/CommandBuilder.php
+++ b/src/PhpBrew/CommandBuilder.php
@@ -75,7 +75,7 @@ class CommandBuilder
         $this->append = $append;
     }
 
-    public function setLogPath($logPath) 
+    public function setLogPath($logPath)
     {
         $this->logPath = $logPath;
     }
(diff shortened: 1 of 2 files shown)


c9s:
@augustohp Thanks for your contribution, could you please also comment the reason why we use `-a` instead of `--apend` in the code? Thanks :+1:

thaJeztah:
> They probably needed to save the memory for some huge background image ..

Thanks for the funny description, it made me :smile:

c9s:
Hahaha Indeed :D

marcioAlmada:
LOL

Well, it looks like `tee -a` is available everywhere so the fix seems legit. I just hope OSX won't eat other long options later :s

related: https://github.com/phpbrew/phpbrew/issues/512

Issue #561, opened 2015-08-17
The clean command is broken


dharmavagabond:
Class [PhpBrew\Command\CleanCommand](https://github.com/phpbrew/phpbrew/blob/4db68e8038620643629e45d29205ad13ec5f0b96/src/PhpBrew/Command/CleanCommand.php#L7) misses `use PhpBrew\Utils`. When I added the import, it threw the following error:

```sh
PHP Fatal error:  Interface 'Universal\ClassLoader\ClassLoader' not found in phar:///Users/Dharma/Desktop/phpbrew/phpbrew/Universal/ClassLoader/SplClassLoader.php on line 2
PHP Stack trace:
PHP   1. {main}() /Users/Dharma/Desktop/phpbrew/phpbrew:0
PHP   2. require() /Users/Dharma/Desktop/phpbrew/phpbrew:3

Fatal error: Interface 'Universal\ClassLoader\ClassLoader' not found in phar:///Users/Dharma/Desktop/phpbrew/phpbrew/Universal/ClassLoader/SplClassLoader.php on line 2

Call Stack:
    0.0002     236080   1. {main}() /Users/Dharma/Desktop/phpbrew/phpbrew:0
    0.0020     408536   2. require('phar:///Users/Dharma/Desktop/phpbrew/phpbrew/Universal/ClassLoader/SplClassLoader.php') /Users/Dharma/Desktop/phpbrew/phpbrew:3
```

jhdxr:
I tried `phpbrew clean` as the documents said, and get the following error:
> Insufficient arguments for command 'clean', which requires 1 arguments, 0 given.

and `phpbrew clean php-5.6.12`
> Makefile not found in path 

I think we should keep the README updated or many people will get confused.

jhdxr:
```
[root@localhost phpbrew]# phpbrew clean php-5.6.12
#0  PhpBrew\Tasks\MakeTask->make(, clean) called at [/root/proj/phpbrew/src/PhpBrew/Tasks/MakeTask.php:26]
[...]

Issue #599, opened 2015-10-21
Allow VersionDslParser to process Alpha/Beta/RC


kohenkatz:
Right now, only release versions can be downloaded by providing the full URL to the download.

The following change to the Regular Expression in the VersionDslParser will also allow downloading Alpha, Beta, and Release Candidate versions:

    (php-(\d.\d+.\d+(?:(?:RC|alpha|beta)\d+)?)\.tar\.(?:gz|bz2))

Confirmed working at: http://www.phpliveregex.com/p/djz

c9s:
Looks godd! Would be great if you can have a PR for this :)

kohenkatz:
@c9s I'll do that.

PR #600, opened 2015-10-22
Feature allow alpha, beta, & RC

kohenkatz:
PR with tests for #599


c9s:
Why delete `.pharcc.yml`?

c9s:
Please reset `phpbrew` phar file from this PR since it will be compiled when rolling out new release

c9s:
If you want to remove `pharcc`, please open another PR to remove. PRs should be simple and contain only one subject

kohenkatz:
I deleted `.pharcc.yml` because CLIFramework 2.8 (which fixes the bug with building the phar) and pharcc aren't compatible with each other because they require conflicting versions of `symfony.yaml`.  I will open a new PR to fix #561 which deletes that old stuff.

c9s:
Thanks!

PR #602, opened 2015-10-22
Update to fixed version of CLIFramework

kohenkatz:
This is a fix for #561. It upgrades `CLIFramework` to the newest version.

In order to do that, some dependencies needed to be updated.  The new `CLIFramework` has an implicit dependency on `symfony/yaml` version 2.7, but the old `pharcc` builder needs version 2.3 of that library.  Because `CLIFramework` has a PHAR builder built into it now, there is no need for the old `pharcc`, so it has been removed.


c9s:
Thanks, let's wait for travis-ci to report the result :)

kohenkatz:
Thanks! How about #601?

PR #609, opened 2015-11-12
Change suggested install location to /usr/local/bin

RobFerrer:
MacOSX 10.11 (El Capitan) introduces extra security that won't let any user (even root) write to /usr, apart from /usr/local. This causes an "Operation not permitted" error when trying to move the executable to /usr/bin.

It seemed to work for me installing to /usr/local/bin, so I've changed the instructions. I don't know if they need changing anywhere else?


c9s:
Looks like 5.3 is failing

Issue #618, opened 2015-11-25
Cannot pass additional arguments to php-fpm in phpbrew fpm start


Ronmi:
I want to run `php-fpm` without daemonize, need to pass `-F` to `php-fpm`. I look into `~/.phpbrew/bashrc` to find out if `phpbrew` has such feature. At line 249, seems I can run `phpbrew fpm start $PHPBREW_PHP -F` to achieve my target, but no success. After dig it further, I found that user arguments does not pass to function `fpm_start` at line 270.

PR #619, opened 2015-11-25
Let user pass additional arguments to php-fpm

Ronmi:
Old code uses wrong way to transfer user arguments, I rewrite it to fit
bash way.

See #618


c9s:
Thanks!

Issue #623, opened 2015-12-03
PHP7 Installation fail with Apache2 patch


Cactucs:
Enable installing PHP 7.

imamassi:
phpbrew install github:php/php-src@PHP-7.0 as php-7.0.0

markwu:
Phpbrew seems only get updates release from php.net for version 5

`https://secure.php.net/releases/index.php?json&version=5&max=100`

markwu:
I got the following errors with `phpbrew install github:php/php-src@PHP-7.0 as php-7.0.0`

```
Build complete.
Don't forget to run 'make test'.

Installing PHP SAPI module:       apache2handler
/usr/local/Cellar/httpd24/2.4.17/build/instdso.sh SH_LIBTOOL='/usr/local/Cellar/apr/1.5.2/libexec/build-1/libtool' libs/libphp7.so /usr/local/Cellar/httpd24/2.4.17/libexec
/usr/local/Cellar/apr/1.5.2/libexec/build-1/libtool --mode=install install libs/libphp7.so /usr/local/Cellar/httpd24/2.4.17/libexec/
libtool: install: install libs/libphp7.so /usr/local/Cellar/httpd24/2.4.17/libexec/libphp7.so
install: libs/libphp7.so: No such file or directory
apxs:Error: Command failed with rc=4653056
.
make: *** [install-sapi] Error 1
```

There is no `libphp.so`, just `libphp7.0.1-dev.so` ... :(

Cactucs:
When I run `phpbrew install github:php/php-src@PHP-7.0 as php-7.0.0` and
the use it I got php version 7.1.0

imamassi:
I don't know the problem here, because it was fine on me, I installed 7.0.0 succesfully

Ronmi:
Anyone willing give my PR a try? Just download [phpbrew from my repository](https://raw.githubusercontent.com/Ronmi/phpbrew/issue623_update_php7_info/phpbrew) and replace original `phpbrew` binary. That works for me.

Leuloch:
I'm using Ubuntu 14.04 and when try:
phpbrew install github:php/php-src@PHP-7.0 as php-7.0.0

I get the following error:
```
configure: error: unrecognized option: `--no-check-certificate'
Try `./configure --help' for more information
```

sebcbi1:
@Ronmi your build is running fine for me. i succesfully installed php 7.0.0 release tarball

[...]

PR #624, opened 2015-12-04
support php7 in phpbrew update

Ronmi:
* Refactoring `buildReleaseListFromOfficialSite()`, move download procedure to a new static method.
* Also download php7 info in `buildReleaseListFromOfficialSite()`.
* Update local release info.
* Update phar.

See #623


assets/php-releases.json
```diff
@@ -1,4 +1,13 @@
 {
+    "7.0": {
```
marcioAlmada: you don't need to update this file, it's deprecated ^^

c9s:
Looks fine, though there is an update for deprecated file.

PR #631, opened 2015-12-09
Fix ReleaseList.php and Options forwarding

Dexus:


Ronmi:
It's my fault that forgetting pass the options, thanks for patching.

Dexus:
you are welcome :) hope a new phpbrew will build soon with this fix...

Ronmi:
@Dexus I just updated phpbrew binary in master branch. Now it works fine with `-o`. Thanks again.

PR #634, opened 2015-12-13
Does not connect the phpbrew bin path when $PHPBREW_BIN is blank

kurotaky:
Hi,
when I was started terminal and I just run `echo $PATH`.
The PATH is displayed as follows.
```
/Users/kurotaky/.phpbrew/php/php-5.6.3/bin::/Users/kurotaky/.plenv/shims:/Users/kurotaky/.rbenv/shims:/Users/kurotaky/.rbenv/shims:/usr/local/Cellar/emacs/HEAD/bin:/Users/kurotaky/.rbenv/bin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/go/bin:/Users/kurotaky/bin
```
Although there is no assigned path to $PHPBREW_BIN, that is connected.
but the empty pathname( like `::`) referred to the current directory in POSIX.
so I fixed problem, the phpbrew bin path is not connected when $PHPBREW_BIN is blank.

PR #640, opened 2015-12-15
fix bin path connection in __phpbrew_set_path()

kurotaky:
Hi, @c9s !
I fixed that the phpbrew bin path is not connected when $PHPBREW_BIN is blank in phpbrew binary file.
related issus: https://github.com/phpbrew/phpbrew/pull/634


marcioAlmada:
Hi,

That diff, or the lack of diff, doesn't look right. What's going on here? :)

marcioAlmada:
Same as https://github.com/phpbrew/phpbrew/pull/641#issuecomment-164626249

You probably want to edit the shell/bashrc file https://github.com/phpbrew/phpbrew/blob/master/shell/bashrc and pull request.

kurotaky:
@marcioAlmada 
Sorry, I just edit the binary file. :cry: 
Thank you for your comment.

PR #641, opened 2015-12-15
Remove unnecessary comments

kurotaky:
I removed unnecessary comments from `shell/phpbrew.fish` and binary files. :)


marcioAlmada:
Hi. We should never edit the binary phpbrew files directly as this will cause phar signature to fail and when a new binary is built the changes would be lost anyway.

Always edit the sources only and do a pull request. The binary will be rebuilt (with `make` command) and pushed by one of the trusted maintainers.

kurotaky:
@marcioAlmada 
OK, I understood it. Thanks!

Issue #577, opened 2015-09-15
Fish fails :(


ovr:
```
Expected a string, but instead found a redirection
~/.phpbrew/phpbrew.fish (line 470):                 set curr_fs (stat -f %d . ^>/dev/null)  # BSD version
                                                                               ^
from sourcing file ~/.phpbrew/phpbrew.fish
	called on line 15 of file ~/.config/fish/config.fish

from sourcing file ~/.config/fish/config.fish
	called during startup
```

Latest master

bcbcarl:
I have the same problem.

sfdi:
Same problem here.

c9s:
I have no idea with fish, @racklin  please save us XD

sfdi:
I think I have a fix here. Try changing line 470 of phpbrew.fish to
```
set curr_fs (stat -f %d . >/dev/null ^&1)
```

racklin:
@c9s  OK!  New version of  Fish-shell 2.10 seen was changed redirection for stderr.

c9s:
The new fish shell script seems fix this issue.

Issue #571, opened 2015-08-28
thoughs about refactoring downloading related


jhdxr:
There are lots of issue about download, especially proxy related ( for example #556 ). and I see this one in the TODO.md
> Ability to switch curl, wget or pure php downloader.

So I think it's time to refactor the download part. 
I'm not sure if it's part of your plan, but if so I'm glad to do this part.

Here are some ideas I planned:
1. refactor `UrlDownloader` to support pure php download (`file_get_contents`?)
2. refactore `UrlDownloader` to add the proxy support for bash command (`wget`/`curl`) and pure php methods (using stream_context)
3. refactore `UrlDownloader` to add the ability to switch the download method.
4. refactore the related codes, using `UrlDownloader` or something based on it to deal with the network request, instead of doing their own downloading everywhere

marcioAlmada:
Hi, I'm not having much time to push fixes and releases lately, but FYI this improvement would be really welcome!

PR #652, opened 2015-12-30
bugfix and improvements for fish

jhdxr:
1. fix issue phpbrew/phpbrew#577
2. add hint for app command


jhdxr:
All the changes I made were based on the branch develop, but I find that develop is behind master a lot. Should I rebased my changes onto master, and create a new pr instead? 

btw, I'm working on #571 currently, and of course, based on develop, is that ok?

c9s:
That is ok :)

PR #656, opened 2016-01-11
Allow user value of variants contains dot character

Gasol:
Because the dot character is prohibited in unix-like filesystem, Also
Homebrew the most popular package manager in OS X that install package
in /usr/local/Cellar, If we prefer use keg-only pkgs to link with PHP
(for example: openssl). We have face the problem that user value can't
contains dot characters. This changes against this issue.


c9s:
Looks good, thanks!

PR #678, opened 2016-02-15
Update readme to reflect most recent known versions

Tuurlijk:


marcioAlmada:
Ty. This is useful so people will know PHP7 is supported.

PR #701, opened 2016-04-02
Network is unreachable php.net fix

zvook:
Hi there!

I've tried to **phpbrew install x.x.x** and get stuck with CURL error **Failed to connect to www.php.net port 80: Network is unreachable**
**tracepath php.net** shows me that one of the path nodes blocks my request to php.net
After some researching I found a problem: node blocks any requests without valid **userAgent** and it's absolutely impossible to install anything using phpbrew

So, I've fixed it by adding default userAgent to the CURL options array
It's very specific issue and cannot be tested by phpUnit, by it solves my problem and now it works like a charm.

### DIFF
--- a/src/PhpBrew/Downloader/UrlDownloader.php
+++ b/src/PhpBrew/Downloader/UrlDownloader.php
@@ -14,6 +14,14 @@ class UrlDownloader
 
     public $options;
 
+    /**
+     * It happens, that some providers blocks curl requests with out user agent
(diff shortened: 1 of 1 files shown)


BeeHuang:
I meet the same problem, but I think is the url error.
http://www.php.net/get/php-5.5.9.tar.bz2/from/this/mirror => error
http://php.net/get/php-5.5.9.tar.bz2/from/this/mirror => correct

Issue #726, opened 2016-05-06
PhpBrew connection timeout


Serginho:
### Output

phpbrew ext install xdebug  
Warning: Some extension won't be able to be built with php7. If the extension
supports php7 in another branch or new major version, you will need to specify
the branch name or version name explicitly.

For example, to install memcached extension for php7, use:

```
phpbrew ext install github:php-memcached-dev/php-memcached php7 -- --disable-memcached-sasl
```

CurlKit\CurlException: (28) Resolving timed out after 10524 milliseconds
Trace:

```
0) CurlKit\CurlDownloader->request('https://pecl.php.net/rest/r/xdebug/stable.txt')
1) PhpBrew\Extension\Provider\PeclProvider->getPackageXml('xdebug', 'stable')
2) PhpBrew\Extension\Provider\PeclProvider->buildPackageDownloadUrl('stable')
3) PhpBrew\Extension\ExtensionDownloader->download(PhpBrew\Extension\Provider\PeclProvider, 'stable')
4) PhpBrew\Command\ExtensionCommand\InstallCommand->execute('xdebug')
5) call_user_func_array([PhpBrew\Command\ExtensionCommand\InstallCommand, 'execute'], ['xdebug'])
6) CLIFramework\CommandBase->executeWrapper(['xdebug'])
7) CLIFramework\Application->run(['/usr/local/bin/phpbrew', 'ext', 'install', 'xdebug'])
8) PhpBrew\Console->runWithTry(['/usr/local/bin/phpbrew', 'ext', 'install', 'xdebug'])
9) require('phar:///usr/local/bin/phpbrew/scripts/phpbrew-emb.php')
```
### Expected Result

No exceptions
### Command

phpbrew ext install xdebug 
### Platform

OS:

```
 Ubuntu 14.04
```

Running PHP: 
 5.5.9 and 7.0.6

I don't know what happens with curl but, I don't have connectivity problems

c9s:
pecl site sometimes down?

Serginho <notifications@github.com> 於 2016年5月6日 星期五寫道：

c9s:
Not sure why you have this timeout, I can't reproduce this.

c9s:
```
phpbrew ext install xdebug                                                                                    !12180
Warning: Some extension won't be able to be built with php7. If the extension
supports php7 in another branch or new major version, you will need to specify
the branch name or version name explicitly.

For example, to install memcached extension for php7, use:

[...]

PR #734, opened 2016-05-09
Feature/pecl downloader

jhdxr:
As mentioned in #726 , `--download` option doesn't apply to `phpbrew ext install` and `phpbrew ext known` when reading info from pecl site.


c9s:
Looks good! Thanks!

PR #781, opened 2016-08-13
Adds psysh.org's php REPL tool

uda:
The `psysh` command is much like `ipython`, works perfectly for interactively running code and debugging.

source: http://psysh.org/


jhdxr:
hello, thanks for your pr. AFAIK, php has a similar tool called phpdbg intergated into the core since 5.6. could you introduce some advantages of psysh comparing to phpdbg (for debug purpose)?

uda:
@jhdxr, These are two different tools, phpdbg is more like gdb for PHP while PsySH is more like IPython, it is an interactive console. I haven't had the chance or need to use the debugging features, so I can't tell how alike/different they are in that field.

PR #784, opened 2016-08-29
Known issues cli fix

lfreneda:
Getting syntax err 

image


jhdxr:
`\` at the end of lines means the command is not over and the next line is appending to the previous line.
so the readme.md is right as it split command into multiple lines.

the printscreen indicates you type everything in one line so you have to remove the `\` to get the correct result.

PR #850, opened 2017-03-12
App store update

peteraba:
Updating  behat3 and box2 versions, fixing behat-3 app name.

### DIFF
--- a/completion/bash/_phpbrew
+++ b/completion/bash/_phpbrew
@@ -1058,12 +1058,12 @@ if [[ $argument_min_length > 0 ]] ; then
 phpunit
 phpmd
 behat-2.5
-behat-3.5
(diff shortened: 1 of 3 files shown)


peteraba:
@c9s please let me know if you think I need to fix something, but I guess the tests are broken.

c9s:
@peteraba looks like it doesn't work for php7

c9s:
could you migrate PHPUnit_Framework_TestCase to `PHPUnit\Framework\TestCase`? because travis ci upgraded phpunit to phpunit 6 I think.

peteraba:
@c9s sorry, I missed your message. I could change that, but:
1. I don't think it has anything to do with my PR
2. It might break other tests I guess.

PR #860, opened 2017-04-07
fix error in fishshell running `phpbrew use xxx`

mcfog:
error message

    [: Expected a combining operator like '-a' at index 1”

in fish `set code (phpbrew env xxx)` set $code as an array instead of
multiline text (see fish-shell/fish-shell#159)

thus `eval $code` become
```
eval "export A=B export C=D"
```

instead of
```
eval "export A=B\
  export C=D"
```

(another problem is the export polyfill function is not robust)

this commit fix that replacing newline with semicol, makes $code a
valid one-liner
```
eval "export A=B;export C=D"
```

also set the exit status (`__phpbrew_set_path ` somehow return nonzero
status)

### DIFF
--- a/shell/phpbrew.fish
+++ b/shell/phpbrew.fish
@@ -126,13 +126,14 @@ function phpbrew
                 set NEW_PHPBREW_PHP_PATH "$PHPBREW_ROOT/php/$_PHP_VERSION"
                 if [ -d $NEW_PHPBREW_PHP_PATH ]
                     if [ $BIN = "phpbrew" ]
-                        set code (command phpbrew env $_PHP_VERSION)
+                        set code (command phpbrew env $_PHP_VERSION | tr '\n' ';')
                     else
-                        set code (eval $BIN env $_PHP_VERSION)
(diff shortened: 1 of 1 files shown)


jhdxr:
Hi @mcfog there is still some cases this pr fail to cover, for example. 
current env is php without curl support, then `phpbrew use 7.0.12` will call `phpbrew env 7.0.12`, and get following return:
```bash
# WARNING: curl extension might be required for fetching data. 
export PHPBREW_ROOT=/root/.phpbrew 
export PHPBREW_HOME=/root/.phpbrew 
export PHPBREW_LOOKUP_PREFIX= 
export PHPBREW_PHP=php-7.0.12 
export PHPBREW_PATH=/root/.phpbrew/php/php-7.0.12/bin 
# Run this command to configure your shell: 
# # eval "$(phpbrew env)"
```

it works in bash becuase only lines start with `#` are ignored, and the export statements still do their job. 

however, your PR changed all the commands into one line starting with `#`, so everything has been ignored. :( 

Could you fix this issue as well?

jhdxr:
merged, thank you.

c9s:
thank you!

PR #897, opened 2017-08-23
fix: Inverted condition for option http-proxy-auth

vincentdaniel:
The condition was inverted.
When no http-proxy-auth was given, the wget was using authentication which resulted in broken http_proxy option such as:
-e http_proxy=@http://my.proxy:8080


jhdxr:
merged, thanks.

PR #937, opened 2017-12-06
Fixing throw of RuntimeException in BaseDownloader

Rickycezar:
One of the thrown exceptions inside the "download" function was being referenced wrongly.


morozov:
Thanks @Rickycezar.

PR #953, opened 2018-02-26
change pecl url from http to https

iliubang:


morozov:
@iliubang please remove the changes in `bin/phpbrew`. This file will be rebuild once a new version is released. Do you also mind changing all `http://*.php.net` URLs for consistency?

iliubang:
@morozov I have removed the changes in `bin/phpbrew`.

morozov:
Thank you. What about the rest of the URLs?

morozov:
@iliubang Why not? There's an HTTPS mirror of http://www.php.net/ at https://secure.php.net/. http://museum.php.net/ is indeed inaccessible via HTTPS.

@jhdxr do you have any objections? Switching to HTTPS means that installing PHP itself or PECL extensions will require PHP compiled with OpenSSL.

jhdxr:
@morozov nope, but I suggest we'd better offer an option for it, like `--ansi` / `--no-ansi`, since sometimes there might be some problem with the local environment during certificate check.

morozov:
@jhdxr do you mean options like `--ssl` / `--no-ssl` which would enable/disable the usage of HTTPS when fetching resources or something like `--no-check-certificate` in `wget` which will disable checking the certificate signature?

The first would make more sense if PHPBrew controlled the web servers where it downloads packages from (like Composer and Packagist). Otherwise, if at some point php.net starts redirecting from HTTP to HTTPs, those switches will become useless.

As for the second, what kind of problems with local environment do you mean? The only valid case I'm aware of when a user wants to disable certificate check is for testing their own service with self-signed certificates. Is there another case when it's needed?

composer.json
```diff
@@ -19,8 +19,8 @@
     "require": {
         "corneltek/cliframework": "3.0.x-dev",
         "corneltek/class-template": "^2",
-        "corneltek/pearx": "dev-master",
-        "corneltek/curlkit": "@dev",
+        "liubang/pearx": "dev-master",
+        "liubang/curlkit": "@dev",
```
c9s: Could you please send the PR to the upstream repo ? thanks.

src/PhpBrew/Console.php
```diff
@@ -14,7 +14,7 @@
 class Console extends Application
 {
     const NAME = 'phpbrew';
-    const VERSION = '1.23.1';
+    const VERSION = '1.23.2';
```
c9s: Could you remove the version change ? thanks

kevinnio:
@iliubang Please rebase `master` so you're Travis build get fixed.

iliubang:
I'm sorry I accidentally deleted the fork,  I have no good way to deal with it now. Could you please delete this PR, or tell me some good idea to handle it.

PR #986, opened 2018-10-28
Use php-vcr to record http interactions

kevinnio:
While submitting #985, I noticed my Travis build kept failing due to API limits while running tests for the `KnownCommand` class.

This introduces `php-vcr` to the project, which allows to record http interactions and reproduce them during tests, thus avoiding the need to actually hit the API. This will solve any API limit issue from now on. Cassettes only need to be updated if the real APIs change.

This PR only makes use of `php-vcr` in `KnownCommandTest` since that's where I noticed most builds fail. If this PR gets accepted, we can start working towards using `php-vcr` for all http interactions in the project's tests.

More info on `php-vcr` at http://php-vcr.github.io.


c9s:
Looks good, thanks.

kevinnio:
Thanks, @c9s. I'll push another PR wrapping all http interactions in the tests with `php-vcr` asap.

PR #987, opened 2018-10-29
Use php-vcr to record http interactions through all the test suite

kevinnio:
This is an attempt to wrap every http interaction done while testing in a VCR cassette. This is a followup to #986.

Not every single interaction can be wrapped since some of them download binary files (which can't be stored in cassettes), but this PR wraps alls interactions that can be wrapped. This will allow for faster builds and less problems related to hitting API limits, triggering anti-DDoS tools and such.

This also introduces two new phpunit group tags: `noVCR` and `macosIncompatible`. The first one is meant to be used in every test which by its own nature cannot be wrapped into a VCR cassette. These are, for example, tests where binary files are downloaded. The latter is just a treat to allow devs who use macOS to run the test suite without tests breaking due to system libraries not available on that OS.

Now we can run something like `vendor/bin/phpunit --exclude-group noVCR` whenever we want to run the test suite without hitting the network. We can always do a full run when needed (while in Travis, for example).

Maybe later we can figure out a way to prevent any test from actually reaching out to the Internet, but I think this is generally a step in a good direction for the project.


kevinnio:
@c9s, if you can take a look when you've got free time, I'd appreciate it.

tests/PhpBrew/Command/InstallCommandTest.php
```diff
@@ -7,16 +7,19 @@
...
  *
- * Don't catch the exceptions, the system command exception 
+ * Don't catch the exceptions, the system command exception
  * will show up the error message.
  *
  * Build output will be shown when assertion failed.
  *
  * @large
  * @group command
+ * @group cantVCR
```
c9s: please rename to `noVCR` for consistency.

c9s: ok, I got it.

c9s:
looks like travis-ci did not run

kevinnio:
Travis build fixed. Ready to merge whenever you are, @c9s.

Issue #1013, opened 2019-05-15
Psysh app url has changed


uda:
### Output

```
CurlKit\CurlException: (22) The requested URL returned error: 404 Not Found at [https://raw.githubusercontent.com/bobthecow/psysh/gh-pages/psysh:The requested URL returned error: 404 Not Found]
```

### Expected Result

Download `psysh` to current php version

### Command

```
phpbrew app get psysh
```

morozov:
Fixed by #1014.

PR #1014, opened 2019-05-15
Fix psysh download url

uda:
The old url `https://git.io/psysh` leads to an old url which no longer contains the latest file.

This change uses the url from the website `https://psysh.org/psysh`.

Solves #1013


uda:
Thanks!

Issue #905, opened 2017-09-13
Issue with sourcing files


MurKit:
When I start my shell, I get:
```
~/.phpbrew/init (line 2): 
Invalid argument system
^
from sourcing file ~/.phpbrew/init
	called on line 224 of file /usr/share/fish/config.fish

in function “.”
	called on line 42 of file ~/.phpbrew/phpbrew.fish

in function “__phpbrew_load_user_config”
	called on line 48 of file ~/.phpbrew/phpbrew.fish

from sourcing file ~/.phpbrew/phpbrew.fish
	called on line 1 of file ~/.config/fish/config.fish

from sourcing file ~/.config/fish/config.fish
	called during startup
```

Linux Mint 17.2

jhdxr:
fish version?

MurKit:
@jhdxr fish, version 2.6.0

jhdxr:
I cannot reproduce this problem (fish 2.6 without third-party theme or plugin, centos, phpbrew 1.23.0)

Issue #926, opened 2017-11-08
phpbrew switch-off Invalid: command not found (but it works)


MurKit:
When switching from phpbrew php-5.3.24 to system 5.6 I got this error:

```
$ phpbrew switch-off
Invalid: command not found
~/.phpbrew/init (line 2): 
Invalid argument 
^
from sourcing file ~/.phpbrew/init
	called on line 224 of file /usr/share/fish/config.fish

in function “.”
	called on line 356 of file ~/.phpbrew/phpbrew.fish

in function “__phpbrew_update_config”
	called on line 367 of file ~/.phpbrew/phpbrew.fish
	with parameter list “”

in function “__phpbrew_reinit”
	called on line 288 of file ~/.phpbrew/phpbrew.fish

in function “phpbrew”
	called on standard input
	with parameter list “switch-off”

phpbrew is switched off.
```

In the end it really switches to system and there are no visible problems.

PR #1015, opened 2019-05-15
Fish shell: fix switch-off and off; add version check for switch

mcfog:
+ switch-off and off not working correctly: see phpbrew/phpbrew#860
+ switch: `phpbrew switch UNEXIST_VERSION` would write "Invalid Argument xxx" into $PHPBREW_HOME/init, thus breaking fish boot procedure.

issues reproduced at (phpbrew 1.23.1 + fish 3.0.2) and (phpbrew 1.22.6 + fish 2.7.1)
fix tested on above environments

maybe alse fixes phpbrew/phpbrew#905 phpbrew/phpbrew#926


c9s:
@racklin I don’t know fish. can you review this fix? Thank you

shell/phpbrew.fish
```diff
@@ -365,7 +372,11 @@ function __phpbrew_reinit
         mkdir -p -p "$PHPBREW_HOME"
     end
     __phpbrew_update_config $_PHP_VERSION
+    if [ $status -ne 0 ]
+        return $status
+    end
     __phpbrew_set_path
+    return 0
```
morozov: This could be omitted for brevity.

shell/phpbrew.fish
```diff
@@ -347,6 +348,12 @@ function __phpbrew_update_config
     else
         set VERSION $argv
     end
+
+    if [ ! -d "$PHPBREW_ROOT/php/$VERSION" ]
+        echo "php version: $VERSION not exists."
```
morozov: Looks like we're missing the same in `bashrc`. Could you add this logic there as well?

c9s:
@mcfog please check the review, thank you

PR #1037, opened 2019-08-08
Allow empty argument to the `env` command.

alex-vlasov:
It was causing 'Invalid argument' error on `switch-off` on fish shell.

On fish shell (v3.0.2) run `phpbrew switch-off`.
The output:
```
Invalid: command not found
~/.phpbrew/init (line 2): 
Invalid argument 
^
from sourcing file ~/.phpbrew/init
	called on line 699 of file ~/.phpbrew/phpbrew.fish

in function “__phpbrew_update_config”
	called on line 726 of file ~/.phpbrew/phpbrew.fish
	with parameter list “”

in function “__phpbrew_reinit”
	called on line 379 of file ~/.phpbrew/phpbrew.fish

in function “phpbrew”
	called on standard input
	with parameter list “switch-off”

phpbrew is switched off.
```
Contents of init file:
```
~ $ cat ~/.phpbrew/init 
# DO NOT EDIT THIS FILE
Invalid argument 
```
Apparently this is caused by fish command `command phpbrew env $VERSION` with empty $VERSION.
While we can trim it in `~/.phpbrew/phpbrew.fish` it's better to make validation less strict and allow empty string.

https://github.com/phpbrew/phpbrew/issues/926 may be related

### DIFF
--- a/src/PhpBrew/Command/EnvCommand.php
+++ b/src/PhpBrew/Command/EnvCommand.php
@@ -16,7 +16,7 @@ class EnvCommand extends \CLIFramework\Command
         $args->add('installed php')
             ->optional()
             ->validValues(function () {
-                return \PhpBrew\Config::getInstalledPhpVersions();
+                return array_merge(\PhpBrew\BuildFinder::findMatchedBuilds(), array(''));
             })
             ;
     }
--- a/tests/PhpBrew/Command/EnvCommandTest.php
+++ b/tests/PhpBrew/Command/EnvCommandTest.php
@@ -25,5 +25,11 @@ class EnvCommandTest extends CommandTestCase
         $this->assertCommandSuccess("phpbrew env");
     }
 
-
+    /**
+     * @outputBuffering enabled
+     */
+    public function testEnvCommandEmptyArg()
+    {
+        $this->assertCommandSuccess("phpbrew env ");
+    }
 }


src/PhpBrew/Command/EnvCommand.php
```diff
@@ -16,7 +16,7 @@ public function arguments($args)
         $args->add('installed php')
             ->optional()
             ->validValues(function () {
-                return \PhpBrew\Config::getInstalledPhpVersions();
+                return array_merge(\PhpBrew\BuildFinder::findMatchedBuilds(), array(''));
```
morozov: The argument is already optional, so adding an empty string as a valid value seems redundant. Furthermore, it will be taken into account during completion which is undesirable.

It looks like the behavior of passing an empty string as an argument is specific to Fish:
```
$ php -r 'echo $_SERVER["argc"];' --
1

$ php -r 'echo $_SERVER["argc"];' -- $UNDEFINED
1

$ set VERSION ""
$ php -r 'echo $_SERVER["argc"];' -- $VERSION
2
```
While Bash behaves in a more expected way:
```
$ VERSION=

$  php -r 'echo $_SERVER["argc"];' -- $VERSION
1
```
But the Fish script doesn't take this behavior into account:

https://github.com/phpbrew/phpbrew/blob/4c93f44deb8fcdec04efecfb8c66c09547d2b37e/shell/phpbrew.fish#L350-L354

tests/PhpBrew/Command/EnvCommandTest.php
```diff
@@ -25,5 +25,11 @@ public function testEnvCommand()
         $this->assertCommandSuccess("phpbrew env");
     }
 
-
+    /**
+     * @outputBuffering enabled
+     */
+    public function testEnvCommandEmptyArg()
+    {
+        $this->assertCommandSuccess("phpbrew env ");
```
morozov: The test doesn't reflect the issue since the trailing space is insignificant and the test won't use any shell to run the command.

PR #1041, opened 2019-09-20
Apply patch to PHP 5.6 for running with OpenSSL 0.11

jasny:
The API of OpenSSL has changed from 0.10 to 0.11. This makes it difficult to run PHP 5 on an updated system. The patch is from https://github.com/php/php-src/pull/2667.

Note that the patch only works for PHP 5.6.31 or higher. While not ideal, being able to install the last (released) PHP5 version will be sufficient for most.

### DIFF
--- a/src/PhpBrew/Patches/PHP56WithOpenSSL11Patch.php
+++ b/src/PhpBrew/Patches/PHP56WithOpenSSL11Patch.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace PhpBrew\Patches;
+
+use PhpBrew\Buildable;
+use PhpBrew\PatchKit\Patch;
+use PhpBrew\PatchKit\DiffPatchRule;
+use CLIFramework\Logger;
+
+class PHP56WithOpenSSL11Patch extends Patch
+{
+    public function desc()
+    {
+        return 'php5.6 with openssl 1.1.x patch.';
+    }
+
+    public function match(Buildable $build, Logger $logger)
+    {
+        return version_compare($build->getVersion(), '5.6') === 0
+            && version_compare($build->getVersion(), '5.6.31') >= 0 // patch only works for 5.6.31 and up
+            && $build->isEnabledVariant('openssl');
+    }
+
+    public function rules()
+    {
+        $rules = array();
+        $rules[] = DiffPatchRule::from('https://patch-diff.githubusercontent.com/raw/php/php-src/pull/2667.patch')
+            ->strip(1)
+            ->sha256('507bb74b2612328cdf5e59b964645a49d0f9367add001ecd698e0fd0d4445421');
+
+        return $rules;
+    }
+}
--- a/src/PhpBrew/Tasks/AfterConfigureTask.php
+++ b/src/PhpBrew/Tasks/AfterConfigureTask.php
@@ -6,6 +6,7 @@ use PhpBrew\Build;
 use PhpBrew\Patches\IntlWith64bitPatch;
 use PhpBrew\Patches\OpenSSLDSOPatch;
 use PhpBrew\Patches\PHP53Patch;
+use PhpBrew\Patches\PHP56WithOpenSSL11Patch;
 
 /**
  * Task run before 'configure'.
@@ -25,6 +26,7 @@ class AfterConfigureTask extends BaseTask
             $patches[] = new PHP53Patch();
             $patches[] = new IntlWith64bitPatch();
             $patches[] = new OpenSSLDSOPatch();
+            $patches[] = new PHP56WithOpenSSL11Patch();
             foreach ($patches as $patch) {
                 $this->logger->info('Checking patch for ' . $patch->desc());
                 if ($patch->match($build, $this->logger)) {


morozov:
@jasny it didn't work for me:
```
$ git log -1 --stat 
commit cfb1ae4c6ecf79edc0b025e7ecba5edde64b6abd (HEAD -> pull/1041)
Author: Arnold Daniels <arnold@jasny.net>
Date:   Fri Sep 20 09:46:13 2019 +0200

    Apply patch to PHP 5.6 for running with OpenSSL 0.11
    The API of OpenSSL has changed from 0.10 to 0.11. This makes it difficult
     to run PHP 5 on an updated system.
    The patch is from https://github.com/php/php-src/pull/2667.

 src/PhpBrew/Patches/PHP56WithOpenSSL11Patch.php | 35 +++++++++++++++++++++++++++++++++++
 src/PhpBrew/Tasks/AfterConfigureTask.php        |  2 ++
 2 files changed, 37 insertions(+)

$ apt policy libssl-dev
libssl-dev:
  Installed: 1.1.1b-1ubuntu2.4
  Candidate: 1.1.1b-1ubuntu2.4
  Version table:
 *** 1.1.1b-1ubuntu2.4 500
        500 http://us.archive.ubuntu.com/ubuntu disco-updates/main amd64 Packages
        100 /var/lib/dpkg/status
     1.1.1b-1ubuntu2 500
        500 http://us.archive.ubuntu.com/ubuntu disco/main amd64 Packages

$ phpbrew --debug install 5.6.39 +neutral +openssl
===> phpbrew will now build 5.6.39
---> Parsing variants from command arguments '+neutral +openssl'
===> Loading and resolving variants...
Checking distribution checksum...
Checksum matched: b3db2345f50c010b01fe041b4e0f66c5aa28eb325135136f153e18da01583ad5
===> Distribution file was successfully extracted, skipping...
Source Directory: /home/morozov/.phpbrew/build/php-5.6.39
Writing variant info to /home/morozov/.phpbrew/php/php-5.6.39/phpbrew.variants
===> Checking patches...
Checking patch for replace apache php module name with custom version name
Enabled variants: [openssl, xml, opcache]
Disabled variants: []
Found existing build.log, renaming it to /home/morozov/.phpbrew/build/php-5.6.39/build.log.1570311882
===> Configuring 5.6.39...


Use tail command to see what's going on:
   $ tail -F /home/morozov/.phpbrew/build/php-5.6.39/build.log


./configure '--cache-file=/home/morozov/.phpbrew/cache/config.cache' '--prefix=/home/morozov/.phpbrew/php/php-5.6.39' '--with-config-file-path=/home/morozov/.phpbrew/php/php-5.6.39/etc' '--with-config-file-scan-dir=/home/morozov/.phpbrew/php/php-5.6.39/var/db' '--with-openssl=/usr' '--enable-dom' '--enable-libxml' '--with-libxml-dir=/usr' '--enable-simplexml' '--enable-xml' '--enable-xmlreader' '--enable-xmlwriter' '--with-xsl' '--enable-opcache' >> /home/morozov/.phpbrew/build/php-5.6.39/build.log 2>&1

Uncaught PhpBrew\Exception\SystemCommandException: Configure failed:  in /home/morozov/Projects/phpbrew/src/PhpBrew/Tasks/ConfigureTask.php on line 107

$ tail /home/morozov/.phpbrew/build/php-5.6.39/build.log
checking for xml2-config path... /usr/bin/xml2-config
checking whether libxml build works... yes
checking for OpenSSL support... yes
checking for Kerberos support... no
checking whether to use system default cipher list instead of hardcoded value... no
checking for DSA_get_default_method in -lssl... no
checking for X509_free in -lcrypto... yes
checking for RAND_egd... no
checking for pkg-config... /usr/bin/pkg-config
configure: error: Cannot find OpenSSL's libraries
```

Are you sure it should be an after-configure task? I'm trying to apply a similar patch in #1044, but it's part of the pre-configure task and requires `./buildconf --force` afterward.

jasny:
@morozov It's tested it, but "works on my machine" :slightly_frowning_face: 

What do you see when you run the following?

    /usr/bin/pkg-config openssl --modversion

Autoconf should be using pkg-config and not even try to find the lib manually.

morozov:
@jasny please look at the log I posted above. The build fails before your patch gets applied. How can it work?

Here's the output:
> /usr/bin/pkg-config openssl --modversion
> 1.1.1b

morozov:
@jasny since #1044 is merged, could you go ahead and use the same approach? You need to move your patch to the pre-configure task and maybe execute the `BuildConfTask` upon a successful patch.

src/PhpBrew/Patches/PHP56WithOpenSSL11Patch.php
```diff
@@ -0,0 +1,35 @@
...
+class PHP56WithOpenSSL11Patch extends Patch
+{
+    public function desc()
+    {
+        return 'php5.6 with openssl 1.1.x patch.';
+    }
+
+    public function match(Buildable $build, Logger $logger)
+    {
+        return true;
```
morozov: This needs to be removed.

jasny:
@morozov I can, but I don't think it will make any difference. The patch doesn't touch the autooconf files and only modifies the c files of the openssl PHP ext.

I can't reproduce the issue you're having by any means. Can you please do some debugging.

In `aclocal.m4` (of php-src) you can find

    AC_DEFUN([PHP_SETUP_OPENSSL],[

In that function you see the section where pkg-config is used to find the openssl libs. It's commented with

    dnl If pkg-config is found try using it
    if test "$PHP_OPENSSL_DIR" = "yes" && test -x "$PKG_CONFIG" && $PKG_CONFIG --exists openssl; then
      if $PKG_CONFIG --atleast-version=0.9.6 openssl; then
        found_openssl=yes
        OPENSSL_LIBS=`$PKG_CONFIG --libs openssl`
        OPENSSL_INCS=`$PKG_CONFIG --cflags-only-I openssl`
        OPENSSL_INCDIR=`$PKG_CONFIG --variable=includedir openssl`
      else
        AC_MSG_ERROR([OpenSSL version 0.9.6 or greater required.])
      fi

      if test -n "$OPENSSL_LIBS"; then
        PHP_EVAL_LIBLINE($OPENSSL_LIBS, $1)
      fi
      if test -n "$OPENSSL_INCS"; then
        PHP_EVAL_INCLINE($OPENSSL_INCS)
      fi
    fi

This should just work. The part where alternative methods are tried should not be executed.

    ...
    dnl If pkg-config fails for some reason, revert to the old method

Please check why `pkg-config` isn't able to find openssl when running `./configure`.

morozov:
@jasny thank you for the clarification. Looks like we're experiencing different issues, and I was under the impression the one you're trying to fix is also related to the configuration, not the code.

I'll follow the steps you suggested above and will provide you the additional information as soon as I can.

morozov:
@jasny, the problem was that when I tried using your build, `--with-libdir=lib/x86_64-linux-gnu` wasn't appended to the `./configure` command. This was recently fixed by #1036. After a rebase, I was able to build PHP:
```
$ php -v
PHP 5.6.39 (cli) (built: Nov  4 2019 12:36:39) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies

$ php -i | grep OpenSSL
OpenSSL support => enabled
OpenSSL Library Version => OpenSSL 1.1.1c  28 May 2019
OpenSSL Header Version => OpenSSL 1.1.1c  28 May 2019
OpenSSL support => enabled
```

src/PhpBrew/Patches/PHP56WithOpenSSL11Patch.php
morozov: Looks like this conversation was resolved mistakenly. We cannot accept this PR in its current state.

danywheeler:
does not work on mac os with brew installed openssl 1.1

markwu:
Weird, this PR seems not working, I have to apply the openssl 1.1 patch manually. Here comes the patch output message:

```
===> Checking patches...
Checking patch for php5.3.29 multi-sapi patch.
Checking patch for php5.3.x on 64bit machine when intl is enabled.
Checking patch for openssl dso linking patch
6 changes patched.
Checking patch for php5.6 with openssl 1.1.x patch.
```

jasny:
@danywheeler The patch is **not** applied when using `phpbrew ext install`, only when installing PHP 5.6 with the openssl variant.

    phpbrew install 5.6.40 +default +openssl --with-openssl="$(brew --prefix openssl)"

_Typically `+openssl` isn't required as it's included in `+default`, but it may be required if you use some other variant like `+minimal`._

---

@markwu Make sure you have the openssl variant (`+openssl`) and are installing PHP 5.6.31 or later. The match is pretty straight forward.

```php
public function match(Buildable $build, Logger $logger)
{
    return version_compare($build->getVersion(), '5.6') === 0
        && version_compare($build->getVersion(), '5.6.31') >= 0 // patch only works for 5.6.31 and up
        && $build->isEnabledVariant('openssl');
}
```

markwu:
Ah ... Okay, I suppose it apply to default. I'll make a test and let you know.

markwu:
I tried again, the rules triggered, but patches nothing.

I have to download the patch and apply it by myself.

#### phpbrew command
```bash
phpbrew --debug install --jobs "$(nproc)" 5.6.40 +default +openssl +dbs +dba +apxs2 +bz2="$(brew --prefix bzip2)" +zlib="$(brew --prefix zlib)" +curl="$(brew --prefix curl)"
```

#### manually patch command
```bash
wget https://patch-diff.githubusercontent.com/raw/php/php-src/pull/2667.patch
patch -p 1 < 2667.patch
```

#### debug log
```
===> Checking patches...
Checking patch for replace apache php module name with custom version name
0 changes patched.
Enabled variants: [openssl, dba, apxs2, bz2, zlib, curl, xml, opcache, sqlite, mysql, pgsql, pdo, bcmath, calendar, cli, ctype, dom, fileinfo, filter, ipc, json, mbregex, mbstring, mhash, mcrypt, pcntl, pcre, pear, phar, posix, readline, sockets, tokenizer, zip]
Disabled variants: []
...
...
===> Checking patches...
Checking patch for php5.3.29 multi-sapi patch.
Checking patch for php5.3.x on 64bit machine when intl is enabled.
Checking patch for openssl dso linking patch
6 changes patched.
Checking patch for php5.6 with openssl 1.1.x patch.
===> Building...
make '-j8' >> /Users/mark/.phpbrew/build/php-5.6.40/build.log 2>&1
Error: Make failed:
The last 5 lines in the log file:
                                                                                 ^

/Users/mark/.phpbrew/build/php-5.6.40/ext/openssl/xp_ssl.c:1840:36: warning: address of 'sslsock->s.timeout' will always evaluate to 'true' [-Wpointer-bool-conversion]

                if (began_blocked && &sslsock->s.timeout) {

                                  ~~  ~~~~~~~~~~~^~~~~~~

12 warnings generated.

Please checkout the build log file for more details:
	 tail /Users/mark/.phpbrew/build/php-5.6.40/build.log
```

markwu:
I have problem with phpbrew 1.24.1 to build php-5.6.40, that's why I revert to phpbrew 1.23.1 and apply this PR only, to  see if I can figure out the problem.

markwu:
Can I get more detail log when do the patch job?

danywheeler:
@jasny Even with the command you typed, it's still not working, don't know what's the problem, but be sure related to openssl. same error as before.

markwu:
Okay, I think I found the problem. the `match` function is not working. I change it to following, evething works!!!

```php
    public function match(Buildable $build, Logger $logger)
    {
        return version_compare($build->getVersion(), '6.0.0') === -1
            && version_compare($build->getVersion(), '5.6.31') >= 0 // patch only works for 5.6.31 and up
            && $build->isEnabledVariant('openssl');
    }

```

@morozov Should I create a new issue for this?

morozov:
@markwu please do. Although, I don't think we should express the requirement for PHP 5.6 using 6.0.0 as a boundary.

markwu:
If so, the only way I can think about is using explode() to find PHP_MAJOR_VERSION.PHP_MINOR_VERSION only, then use the version string to compare.

morozov:
@markwu indeed. Please go ahead and use your original approach. Given how non-obvious this code is, you can try adding a test that covers the `match()` method of the patch.

Issue #872, opened 2017-05-09
Add PHPBREW_SYSTEM_PHP variable to run the system PHP for phpbrew


c9s:

morozov:
In https://github.com/phpbrew/phpbrew/issues/988#issuecomment-559942046, @theofidry said:

> It won't solve the problem if one user try to install phpbrew on a system which doesn't have any recent version of PHP. But I guess this could be mitigated by providing a cask recipe or docker image...

Right. There are two different use cases for PHPBrew::

1. Building an older PHP version on a newer OS that ships with a newer PHP version.
2. Building a newer PHP version on an older OS that ships with an older PHP version.

For me, it was always #‌1 but not #‌2. The reason is: I consider PHP a development tool. As a developer who's capable of maintaining my own development environment up to date, I want to develop or maintain an old PHP application that needs an old PHP version. So I compile, configure and manage it with PHPBrew. Therefore, I think we can always require the oldest supported version of PHP since it will be available on the developer's OS which the developer is in charge of maintaining.

What is the rationale behind #‌2?

theofidry:
The machine does not provide the required version and phpbrew is a useful tool to install PHP :)

morozov:
That's the question. Why doesn't the machine provide even the oldest supported version of PHP? Could you provide a real example?

theofidry:
Because some distributions are lagging behind,  such as Debian or CenterOs.
That said if you don’t think it’s a big deal then it’s fine to ignore the
problem for now. I by no means imply this is a blocker

morozov:
This is true. I'm using Ubuntu 16.04 right now which ships with PHP 7.0. The distro is still supported by Canonical but the PHP is no longer supported by its community.

One of the solutions is that we can suggest using an older PHPBrew version, compile the latest PHP version and then upgrade. Or, alternatively, we can identify the list of older distros we want to support and come up with super simple shell scripts or instructions of building the system PHP runtime there in order to be able to run PHPBrew and build everything else.

Another solution might be to use a 3rd party distribution channel like [Ondřej Surý's PPA](https://launchpad.net/~ondrej/+archive/ubuntu/php) in a similar way.

Ideas are welcome.

PR #1045, opened 2019-10-14
Update .travis.yml

andreybolonin:


morozov:
@andreybolonin it looks like it will require to migrate to xenial distribution first.

jhdxr:
I tried to migrate a very long time ago but failed. I doubt xenial, or any other new environment works for old versions like 5.3.

morozov:
That's fair. I don't think any existing environment will support the range of PHP versions we support. Migration to a newer environment would be possible after https://github.com/phpbrew/phpbrew/issues/872 is resolved and we drop the support of older PHP versions as runtime. I hope I can build a draft for that somewhere during the winter holidays.

PR #1050, opened 2019-11-04
deprecated mirror option since php.net retired the mirror program

jhdxr:
ref: https://externals.io/message/104533


morozov:
Thank you @jhdxr.

PR #1064, opened 2019-12-03
Added `sodium` as variant.

jasny:
Sodium is part of the PHP core since 7.2.
See https://www.php.net/manual/en/sodium.installation.php

Added `sodium` to the `default` virtual variant.

### DIFF
--- a/src/PhpBrew/VariantBuilder.php
+++ b/src/PhpBrew/VariantBuilder.php
@@ -139,6 +139,7 @@ class VariantBuilder
             'posix',
             'readline',
             'sockets',
+            'sodium',
             'tokenizer',
             'xml',
             'curl',
@@ -548,6 +549,27 @@ class VariantBuilder
         $this->variants['icu'] = function (Build $build, $val = null) {
             if ($val) {
                 return '--with-icu-dir=' . $val;
+            }
+        };
+
+        $this->variants['sodium'] = function (Build $build, $prefix = null) {
+            if ($build->compareVersion('7.2') < 0) {
+                echo "Sodium is available as a core extension since PHP 7.2.0. Please use 'phpbrew ext install sodium' to install it from PECL\n";
+            }
+
+            if ($prefix === null) {
+                $prefix = Utils::findPrefix(array(
+                    new BrewPrefixFinder('libsodium'),
+                    new PkgConfigPrefixFinder('libsodium'),
+                    new IncludePrefixFinder('sodium.h'),
+                    new LibPrefixFinder('libsodium.a'),
+                ));
+            }
+
+            if ($prefix !== null) {
+                return '--with-sodium=' . $prefix;
+            } else {
+                return '--with-sodium';
             }
         };
 


src/PhpBrew/VariantBuilder.php
```diff
@@ -561,6 +562,39 @@ public function __construct()
...
+         * --with-sodium=[dir]
+         *
+         * On ubuntu you need to install libsodium-dev
+         */
+        $this->variants['sodium'] = function (Build $build, $val = null) {
+            if ($val) {
+                return "--with-sodium=$val";
+            }
+
+            if ($prefix = Utils::findIncludePrefix('sodium.h')) {
```
morozov: Could you please try using prefix finder like it's done for `libxml`?

https://github.com/phpbrew/phpbrew/blob/01de86c80079696899782c89922f8d69dbc0b9dd/src/PhpBrew/VariantBuilder.php#L760-L769

This way, the code is more expressive and is easier to maintain.

src/PhpBrew/VariantBuilder.php
```diff
@@ -561,6 +562,39 @@ public function __construct()
             }
         };
 
+        /*
+         * --with-sodium option
+         *
+         * --with-sodium=shared
```
morozov: This is misleading. The implementation doesn't configure it as shared.

src/PhpBrew/VariantBuilder.php
```diff
@@ -561,6 +562,39 @@ public function __construct()
...
+            }
+
+            $possiblePrefixes = array('/usr/local/opt/sodium');
+            $foundPrefixes = array_filter($possiblePrefixes, 'file_exists');
+            if (count($foundPrefixes) > 0) {
+                return '--with-sodium='.$foundPrefixes[0];
+            }
+
+            // This will create sodium.so file for dynamic loading.
+            echo 'Compiling with sodium=shared, please install libsodium-dev or sodium header files if you need';
```
morozov: This is misleading. The implementation doesn't configure it as shared. Also, the message about dependencies doesn't cover all supported OS'es and is prone to become invalid over time, so it's better to move it into [Wiki](https://github.com/phpbrew/phpbrew/wiki/Variant-Dependencies).

src/PhpBrew/VariantBuilder.php
```diff
@@ -561,6 +562,39 @@ public function __construct()
             }
         };
 
+        /*
+         * --with-sodium option
```
morozov: This is an implementation detail that may change for a newer PHP version. Please remove.

src/PhpBrew/VariantBuilder.php
```diff
@@ -561,6 +562,39 @@ public function __construct()
...
+         *
+         * --with-sodium=shared
+         * --with-sodium=[dir]
+         *
+         * On ubuntu you need to install libsodium-dev
```
morozov: This should go to Wiki.

src/PhpBrew/VariantBuilder.php
```diff
@@ -561,6 +562,39 @@ public function __construct()
...
+
+            if ($prefix = Utils::findIncludePrefix('sodium.h')) {
+                return "--with-sodium=$prefix";
+            }
+
+            if ($prefix = Utils::getPkgConfigPrefix('sodium')) {
+                return "--with-sodium=$prefix";
+            }
+
+            $possiblePrefixes = array('/usr/local/opt/sodium');
```
morozov: If this is expected to be created by HomeBrew, we should use the `BrewPrefixFinder` instead of hard-coding the path.

src/PhpBrew/VariantBuilder.php
```diff
@@ -551,6 +552,26 @@ public function __construct()
             }
         };
 
+        $this->variants['sodium'] = function (Build $build, $val = null) {
```
morozov: The user-provided prefix probably shouldn't be ignored. Look at:
https://github.com/phpbrew/phpbrew/blob/b2171dc8bc4b4de7cfde39b34719cb337f7454d2/src/PhpBrew/VariantBuilder.php#L418-L420

src/PhpBrew/VariantBuilder.php
```diff
@@ -551,6 +552,26 @@ public function __construct()
...
+            if ($build->compareVersion('7.2') < 0) {
```
morozov: Shouldn't the user just not include the variant that is not supported by a given PHP version? Otherwise, if a process of building PHP is automated, the user would expect to get the extension available upon successful completion but it won't be true if we just omit this variant for PHP 7.1 and older.

On the other hand, if sodium is unpacked under the `ext` directory and `phpize`d, it probably may work even with older PHPs, so PHPBrew shouldn't prevent that.

jasny: In PHP 7.1 and older, sodium isn't bundled with PHP and `--with-sodium` isn't recognized by configure. If you use `--with-sodium` with older PHP versions the build will just succeed, but you won't have sodium. Removing this check doesn't solve the issue of automated builds you're describing.

Without the warning, the only indication that something went wrong is a line the build log

    configure: WARNING: unrecognized options: --with-sodium

This typically goes unnoticed.

jasny: Manually adding extensions to the build dir and using variants to include them isn't supported by phpbrew. Variants can only be used for extensions that are bundled. I don't see why this should be the exception.

morozov: Then this is just a replacement of one more or less standard warning with another warning specific to PHPBrew. It's better if we keep the standard one which is easier to google and understand what's going on.

jasny: This warning is given on-screen, not just in the log. It's similar to warnings for other extensions like

    echo "homebrew prefix '$output' doesn't exist. you forgot to install?\n";

I did just remove the `return`. In other cases, the option is added regardless of the warning. Here it's now the same.

morozov: Yeah, that should be fine.

src/PhpBrew/VariantBuilder.php
```diff
@@ -551,6 +552,28 @@ public function __construct()
             }
         };
 
+        $this->variants['sodium'] = function (Build $build, $prefix = null) {
+            if ($build->compareVersion('7.2') < 0) {
+                echo "Sodium is builded since PHP version 7.2.0, please install sodium as extension.\n";
+                return;
```
morozov: > I did just remove the `return`.

It's still here.

jasny: Oops, forgot to push.

src/PhpBrew/VariantBuilder.php
```diff
@@ -551,6 +552,27 @@ public function __construct()
             }
         };
 
+        $this->variants['sodium'] = function (Build $build, $prefix = null) {
+            if ($build->compareVersion('7.2') < 0) {
+                echo "Sodium is builded since PHP version 7.2.0, please install sodium as extension.\n";
```
morozov: Sorry for the last minute request but this needs to be changed. "Builded" is not a word in English.

How about "Sodium is available as a core extension since PHP 7.2.0. Please use 'phpbrew ext install sodium' to install it from PECL"?

jasny: I meant to say 'bundled' instead of 'builded'. I've changed it to your suggestion.

morozov: I thought about it this way but "bundled" in this context means that the underlying C library (libsodium) is packaged along the PHP sources like it used to be done with libzip (see https://github.com/php/php-src/pull/1828) which is not true about libsodium and is not relevant in this case.

morozov:
Thank you @jasny. If you know which dependencies the `sodium` variant has on OS'es other than Ubuntu, please update the [wiki page](https://github.com/phpbrew/phpbrew/wiki/Variant-Dependencies#sodium).

Issue #988, opened 2018-10-30
Project status


theofidry:
Hello there. 

I find that project very helpful and love to use it. However, its maintenance seems a bit sporadic. I wonder if anything else could be done to help with that? From what I see there seem to be a few problematic points:

- @c9s availability: maybe more contributors could shim in as maintainers to help out and avoid c9s to be too pressured and be the contribution bottleneck.
- The insane old dependencies: PHPBrew is compatible with _PHP 5.3_. In itself, it's perfectly fine as it allows to broaden the range of users like Composer does. However although quite popular, PHPBrew is not nearly as used, popular and well maintained as Composer and its codebase and range of features is also significantly bigger. This begs the question: should the minimal requirements be updated? I think a better solution would be to upgrade to PHP 7+ to help out with the codebase. If PHP 5.x support is still necessary, maybe solutions to transpile new PHP code to older versions such as [7to5](https://github.com/spatie/7to5) should be considered. For example, this could take the form of a [Box compactor](https://github.com/humbug/box/blob/master/doc/configuration.md#compactors-compactors) so that the shipped PHPBrew PHAR remains PHP5.3 compatible whilst the codebase can happily benefit from newer PHP features.
- The lack of visibility in the work left to do. There is a bunch of issues all over the place but overall I think maybe phpbrew should cut down some features such as the App Store in favor eventually of a [phive](https://github.com/phar-io/phive) integration. In other words, we are missing a kind of roadmap to really help out with the project as contributors.

Thoughts?

/cc @c9s

jasny:
Second that. I'm more than happy to spend some time on the project. However, I'm a bit reluctant to do so if PRs just stay open. The project has become too successful and important as a single person side project. Adding maintainers seem like a must to keep the project thriving.

theofidry:
cc @jhdxr which apparently has writing access to the repo as well :)

jhdxr:
As for the second point, I agree we increase the minimum requirements to 7+ when php 5.6 reach its EOL. and of course we should also increase the major version of phpbrew for this minimum requirements update. we may also consider to maintain the old branch for a period of time, but at some point we may finally drop it.
Reason behind this decision: The latest OS may not be able to compile the old PHP (travis-ci drops php 5.3 on Ubuntu Trusty), and old OS may not be able to compile the latest PHP as well. So one day, we may be in a dilemma that although phpbrew supported all versions, php itself wouldn't be able to be compiled.

jhdxr:
[...]

PR #1068, opened 2019-12-06
Remove AppStore

theofidry:
See the original discussion in #988.


theofidry:
Hm, I don't quite get how that works I guess :/

morozov:
> Hm, I don't quite get how that works I guess :/

What do you mean? From the test failures, it seems like that there are some AppStore code leftovers. But besides that, I'd first like to deprecate this functionality and then remove it. Removing it just like that might be too harsh.

jhdxr:
>I'd first like to deprecate this functionality and then remove it. 

Agreed. Removing a future in a minor upgrade without any deprecation is really a bad idea IMO. 

btw, could you provide some justification for this removal?

jhdxr:
ok. I got why you propose this PR. but I still prefer to remove it in a major upgrade. (maybe someone could help me to remove the deprecated --mirror option at the same time)

theofidry:
Because there is tool such as Phive that aren’t both more appropriate and
complete :)

morozov:
> but I still prefer to remove it in a major upgrade

Given the current lack of users and maintainers, it might be too much burden to adhere to semantic versioning. The way I see it right now is: fix regressions in patch releases, add new features, deprecate and remove the old ones in minor releases. At the current state, a major release is out of the question, IMO.

> maybe someone could help me to remove the deprecated --mirror option at the same time

Removed in #1083.

morozov:
#1088 is filed to deprecate the component first.

theofidry:
Let's close this one in favour of #1088 then :)

Issue #1040, opened 2019-09-12
php 7.4 on macos


if4lcon:
Alternatively, you may set the environment variables OPENSSL_CFLAGS

and OPENSSL_LIBS to avoid the need to call pkg-config.

See the pkg-config man page for more details.

no matter what command I changed I even copied php 7.3 configure parameters from homebrew

bskiefer:
Not sure if this helps in your case, but my fix was to add the following to my `.bashrc`/`.zshrc` file.

```
CPPFLAGS=-I/usr/local/opt/openssl/include:"${CPPFLAGS}"  
LD_LIBRARY_PATH=/usr/local/Cellar/openssl/1.0.2s/lib:"${LD_LIBRARY_PATH}"
CPATH=/usr/local/opt/openssl/include:"${CPATH}"  
PKG_CONFIG_PATH=/usr/local/Cellar/openssl/1.0.2s/lib/pkgconfig:"${PKG_CONFIG_PATH}"          
export LD_LIBRARY_PATH CPATH PKG_CONFIG_PATH CPPFLAGS
```

May have to adjust as necessary, but as you can tell I'm using brew's openssl for pkg config & LD.

if4lcon:
thank you ... unfortunately didn't compile

```
Alternatively, you may set the environment variables SASL_CFLAGS

and SASL_LIBS to avoid the need to call pkg-config.

See the pkg-config man page for more details.```

**Comment by ramsey** [comment]
I'm having the same problem. It looks like phpbrew can no longer find the Homebrew libraries.

**Comment by bskiefer** [comment]
@ramsey What error are you getting? It's sometimes useful to run `brew link {{package}} --force`, or reinstalling the dependency. However I had to liter my .zshrc with exports to get everything to work correctly.

I'll just include a few that solved my issues

```

export CPPFLAGS=-I/usr/local/opt/openssl/include:"${CPPFLAGS}"  
export CPPFLAGS=-I/usr/local/opt/libedit/include:"${CPPFLAGS}"  
export CPPFLAGS=-I/usr/local/Cellar/readline/8.0.1/include:"${CPPFLAGS}"  

export LD_LIBRARY_PATH=/usr/local/Cellar/openssl/1.0.2s/lib:"${LD_LIBRARY_PATH}"
export LD_LIBRARY_PATH=/usr/local/opt/libedit/lib:"${LD_LIBRARY_PATH}"  
[...]

PR #1085, opened 2019-12-10
Add PKG_CONFIG_PATH environment variables to phpbrew build process

markwu:
I try to dig the source code and add `PKG_CONFIG_PATH` to build process, it makes easier for user when build php 7.4.

It is just  a proof of concept, for discussion only.

With the PR, you can build php 7.4, with the following command only, no more `export PKG_CONFIG_PATH=...`.
```bash
$ phpbrew --debug install --jobs "$(nproc)" 7.1.33 +default +dbs +dba +apxs2 +bz2="$(brew --prefix bzip2)" +zlib="$(brew --prefix zlib)" +ftp
```
But, there is a problem when build extensions, take `ftp` and `soap` for example.  You can install it with `phpbrew install 7.4.0 +default +ftp`, the build process will set the PKG_CONFIG_PATH for you.

But, with this PR, you still can't install extension by following command only.
```
phpbrew ext install ftp
phpbrew ext install soap
```
Because the current extension manager install them by `phpize`, `configure`, `make`. Without the capability like `VariantBuilder` add extra settings for you...

So, You still need to export `PKG_CONFIG_PATH`, `LDFLAGS`, `CPPFLAGS` manually. Then, install the extension.

@morozov Need your review and ideas.

Reference: https://github.com/phpbrew/phpbrew/issues/1040#issuecomment-562501149

### DIFF
--- a/src/PhpBrew/Build.php
+++ b/src/PhpBrew/Build.php
@@ -61,6 +61,13 @@ class Build implements Serializable, Buildable
     public $osRelease;
 
     /**
+     * PKG_CONFIG_PATH for build environment
+     *
+     * @var array<string,true>
+     */
+    private $pkgConfigPaths = array();
+
+    /**
      * Construct a Build object,.
      *
      * A build object contains the information of all build options, prefix, paths... etc
@@ -81,6 +88,28 @@ class Build implements Serializable, Buildable
         $this->setBuildSettings(new BuildSettings());
         $this->osName = php_uname('s');
         $this->osRelease = php_uname('r');
+    }
+
+    public function getPkgConfigPaths()
+    {
+        return array_keys($this->pkgConfigPaths);
+    }
+
+    public function addPkgConfigPath($path)
+    {
+        $this->pkgConfigPaths[$path] = true;
+    }
+
+    public function putPkgConfigPathsEnv()
+    {
+        $paths = $this->getPkgConfigPaths();
+        $currentPkgConfigPath = getenv('PKG_CONFIG_PATH');
+
+        if ($currentPkgConfigPath !== '' && $currentPkgConfigPath !== false) {
+            $currentPaths = explode(PATH_SEPARATOR, $currentPkgConfigPath);
+            $paths = array_unique(array_merge($paths, $currentPaths));
+        }
+        putenv('PKG_CONFIG_PATH=' . implode(PATH_SEPARATOR, $paths));
     }
 
     public function setName($name)
--- a/src/PhpBrew/Command/InstallCommand.php
+++ b/src/PhpBrew/Command/InstallCommand.php
@@ -371,6 +371,9 @@ class InstallCommand extends Command
         $this->logger->debug('Source Directory: ' . realpath($targetDir));
         $build->setSourceDirectory($targetDir);
 
+        // Update PKG_CONFIG_PATH gathered from VariantBuilder to build environment
+        $build->putPkgConfigPathsEnv();
+
         if (!$this->options->{'no-clean'} && file_exists($targetDir . DIRECTORY_SEPARATOR . 'Makefile')) {
             $this->logger->info('Found existing Makefile, running make clean to ensure everything will be rebuilt.');
             $this->logger->info(
--- a/src/PhpBrew/VariantBuilder.php
+++ b/src/PhpBrew/VariantBuilder.php
@@ -295,44 +295,47 @@ class VariantBuilder
             return '--with-mcrypt'; // let autotool to find it.
         };
 
-        $this->variants['zlib'] = function (Build $build, $prefix = null) {
-            if ($prefix) {
-                return "--with-zlib=$prefix";
-            }
-            if ($prefix = Utils::findIncludePrefix('zlib.h')) {
-                return "--with-zlib=$prefix";
-            }
+        $this->variants['zlib'] = function (Build $build, $val = null) {
+            $prefix = Utils::findPrefix(array(
+                new BrewPrefixFinder('zlib'),
+                new IncludePrefixFinder('zlib.h'),
+            ));
 
-            return;
-        };
+            if ($build->compareVersion('7.4') < 0) {
+                if ($val) {
+                    return '--with-zlib=' . $val;
+                }
 
-        $this->variants['curl'] = function (Build $build, $prefix = null) {
-            if ($prefix) {
-                return "--with-curl=$prefix";
-            }
-
-            if ($prefix = Utils::findIncludePrefix('curl/curl.h')) {
-                return "--with-curl=$prefix";
-            }
-
-            if ($prefix = Utils::getPkgConfigPrefix('libcurl')) {
-                return "--with-curl=$prefix";
-            }
-
-            if ($bin = Utils::findBin('brew')) {
-                if ($prefix = exec_line("$bin --prefix curl")) {
-                    if (file_exists($prefix)) {
-                        return "--with-curl=$prefix";
-                    }
-                    echo "homebrew prefix '$prefix' doesn't exist. you forgot to install?\n";
+                if ($prefix !== null) {
+                    return '--with-zlib=' . $prefix;
+                }
+            } else {
+                if ($prefix !== null) {
+                    $build->addPkgConfigPath($prefix . '/lib/pkgconfig');
                 }
             }
-            if ($bin = Utils::findBin('curl-config')) {
-                if ($prefix = exec_line("$bin --prefix")) {
-                    if (file_exists($prefix)) {
-                        return "--with-curl=$prefix";
-                    }
-                    echo "homebrew prefix '$prefix' doesn't exist. you forgot to install?\n";
+
+            return '--with-zlib';
+        };
+
+        $this->variants['curl'] = function (Build $build, $val = null) {
+            $prefix = Utils::findPrefix(array(
+                new BrewPrefixFinder('curl'),
+                new PkgConfigPrefixFinder('libcurl'),
+                new IncludePrefixFinder('curl/curl.h'),
+            ));
+
+            if ($build->compareVersion('7.4') < 0) {
+                if ($val) {
+                    return '--with-curl=' . $val;
+                }
+
+                if ($prefix !== null) {
+                    return '--with-curl=' . $prefix;
+                }
+            } else {
+                if ($prefix !== null) {
+                    $build->addPkgConfigPath($prefix . '/lib/pkgconfig');
                 }
             }
 
@@ -583,37 +586,25 @@ class VariantBuilder
          * On ubuntu you need to install libssl-dev
          */
         $this->variants['openssl'] = function (Build $build, $val = null) {
-            if ($val) {
-                return "--with-openssl=$val";
-            }
+            $prefix = Utils::findPrefix(array(
+                new BrewPrefixFinder('openssl'),
+                new PkgConfigPrefixFinder('openssl'),
+                new IncludePrefixFinder('openssl/opensslv.h'),
+            ));
 
-            if ($prefix = Utils::findIncludePrefix('openssl/opensslv.h')) {
-                return "--with-openssl=$prefix";
-            }
+            if ($build->compareVersion('7.4') < 0) {
+                if ($val) {
+                    return '--with-openssl=' . $val;
+                }
 
-            // Special detection and fallback for homebrew openssl
-            // @see https://github.com/phpbrew/phpbrew/issues/607
-            if ($bin = Utils::findBin('brew')) {
-                if ($output = exec_line("$bin --prefix openssl")) {
-                    if (file_exists($output)) {
-                        return "--with-openssl=$output";
-                    }
-                    echo "prefix $output doesn't exist.";
+                if ($prefix !== null) {
+                    return '--with-openssl=' . $prefix;
+                }
+            } else {
+                if ($prefix !== null) {
+                    $build->addPkgConfigPath($prefix . '/lib/pkgconfig');
                 }
             }
-
-            if ($prefix = Utils::getPkgConfigPrefix('openssl')) {
-                return "--with-openssl=$prefix";
-            }
-
-            $possiblePrefixes = array('/usr/local/opt/openssl');
-            $foundPrefixes = array_filter($possiblePrefixes, 'file_exists');
-            if (count($foundPrefixes) > 0) {
-                return '--with-openssl=' . $foundPrefixes[0];
-            }
-
-            // This will create openssl.so file for dynamic loading.
-            echo 'Compiling with openssl=shared, please install libssl-dev or openssl header files if you need';
 
             return '--with-openssl';
         };
@@ -767,21 +758,25 @@ class VariantBuilder
                 '--enable-dom',
             );
 
-            if ($build->compareVersion('7.4') < 0) {
-                $options[] = '--enable-libxml';
-
-                if (
-                    ($prefix = Utils::findPrefix(array(
+            $prefix = Utils::findPrefix(array(
                     new BrewPrefixFinder('libxml2'),
                     new PkgConfigPrefixFinder('libxml'),
                     new IncludePrefixFinder('libxml2/libxml/globals.h'),
                     new LibPrefixFinder('libxml2.a'),
-                    ))) !== null
-                ) {
+            ));
+
+            if ($build->compareVersion('7.4') < 0) {
+                $options[] = '--enable-libxml';
+
+                if ($prefix !== null) {
                     $options[] = '--with-libxml-dir=' . $prefix;
                 }
             } else {
                 $options[] = '--with-libxml';
+
+                if ($prefix !== null) {
+                    $build->addPkgConfigPath($prefix . '/lib/pkgconfig');
+                }
             }
 
             $options = array_merge($options, array(


src/PhpBrew/Command/InstallCommand.php
```diff
@@ -343,6 +343,8 @@ public function execute($version)
         $variantBuilder = new VariantBuilder();
         $configureOptions = $variantBuilder->build($build);
 
+        $build->putEnvVars();
```
morozov: Why does having the build object properly initialized require calling an extra method on it? Can it be done by default?

src/PhpBrew/VariantBuilder.php
```diff
@@ -315,34 +315,27 @@ public function __construct()
...
 
-        $this->variants['curl'] = function (Build $build, $prefix = null) {
-            if ($prefix) {
-                return "--with-curl=$prefix";
-            }
-
-            if ($prefix = Utils::findIncludePrefix('curl/curl.h')) {
-                return "--with-curl=$prefix";
-            }
+        $this->variants['curl'] = function (Build $build, $val = null) {
```
morozov: Why is the argument changed from `$prefix` to `$val`?

src/PhpBrew/VariantBuilder.php
```diff
@@ -315,34 +315,27 @@ public function __construct()
             return;
         };
 
-        $this->variants['curl'] = function (Build $build, $prefix = null) {
-            if ($prefix) {
```
morozov: Why is this condition removed? What's the point of looking up the prefix if it's passed explicitly?

src/PhpBrew/VariantBuilder.php
```diff
@@ -315,34 +315,27 @@ public function __construct()
...
+            $prefix = Utils::findPrefix(array(
+                new BrewPrefixFinder('curl'),
+                new PkgConfigPrefixFinder('curl'),
+                new IncludePrefixFinder('curl/curl.h'),
+            ));
 
-            if ($prefix = Utils::getPkgConfigPrefix('libcurl')) {
-                return "--with-curl=$prefix";
-            }
+            if ($build->compareVersion('7.4') < 0) {
```
morozov: The `+curl` variant seems to work fine on Linux regardless of the PHP version. Is it specific to macOS and what exactly has changed in PHP in that regard?

src/PhpBrew/Build.php
```diff
@@ -98,6 +100,31 @@ public function setOSRelease($osRelease)
         $this->osRelease = $osRelease;
     }
 
+    public function setEnvVar($key, $value)
```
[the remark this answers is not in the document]
markwu: Thanks for this detail review. As I said, it's a proof of concept. Yes, I agree `setPkgConfigPath($path)` is more clear.

src/PhpBrew/Build.php
```diff
@@ -98,6 +100,31 @@ public function setOSRelease($osRelease)
...
+    }
+
+    public function putEnvVars()
+    {
+        foreach ($this->envVars as $key => $elements) {
+            if ($key === 'PKG_CONFIG_PATH') {
+                $value = implode(PATH_SEPARATOR, $elements);
+                $pkgConfigPath = getenv('PKG_CONFIG_PATH');
+                if ($pkgConfigPath === '') {
+                    putenv('PKG_CONFIG_PATH='.$value);
```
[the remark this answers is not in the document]
markwu: I am sure how to do it, that's why I put it in current process to proof it's doable.

src/PhpBrew/Build.php
```diff
@@ -98,6 +100,31 @@ public function setOSRelease($osRelease)
...
+                } else {
+                    putenv('PKG_CONFIG_PATH='.$pkgConfigPath.PATH_SEPARATOR.$value);
+                }
+            }
+        }
+    }
+
+    public function getEnvVars()
```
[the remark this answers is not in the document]
markwu: Yes, I can remove it.

src/PhpBrew/Build.php
```diff
@@ -65,6 +65,8 @@ class Build implements Serializable, Buildable
 
     public $osRelease;
 
+    public $envVars = array();
```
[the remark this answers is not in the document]
markwu: I just follow variables in this class, that's why I make it `public`.

src/PhpBrew/Command/InstallCommand.php
markwu: I am not sure, I will try to do it.

src/PhpBrew/VariantBuilder.php
markwu: I did not remove it, I just follow the `libxml2` you modified, and change the name to `$val`. And move it to the following statements
```php
            if ($build->compareVersion('7.4') < 0) {
                if ($val) {
                    return "--with-curl=$val";
                }
            ...
```

markwu:
The logic of `openssl` and `curl` just follow the pattern you modified in `libxml2`,

I try to convert the logic current to the following, and remain the same

1. lookup `$prefix` through  `Util::findPrefix` with finders.
```php
            $prefix = Utils::findPrefix(array(
                new BrewPrefixFinder('libxml2'),
                new PkgConfigPrefixFinder('libxml'),
                new IncludePrefixFinder('libxml2/libxml/globals.h'),
                new LibPrefixFinder('libxml2.a'),
            ));
```
```php
            $prefix = Utils::findPrefix(array(
                new BrewPrefixFinder('curl'),
                new PkgConfigPrefixFinder('curl'),
                new IncludePrefixFinder('curl/curl.h'),
            ));
```

2. Then, I check the version is older or newer than `7.4`

```php
            if ($build->compareVersion('7.4') < 0) {
                $options[] = '--enable-libxml';

                if ($prefix !== null) {
                    $options[] = '--with-libxml-dir=' . $prefix;
                }
            } else {
                $options[] = '--with-libxml';

                if ($prefix !== null) {
                    $build->setEnvVar('PKG_CONFIG_PATH', $prefix.'/lib/pkgconfig');
                }
            }

            $options = array_merge($options, array(
                '--enable-simplexml',
                '--enable-xml',
                '--enable-xmlreader',
                '--enable-xmlwriter',
                '--with-xsl',
            ));

            return $options;
        };
```
```php
            if ($build->compareVersion('7.4') < 0) {
                if ($val) {
                    return "--with-curl=$val";
                }

                if ($prefix !== null) {
                    return '--with-curl=' . $prefix;
                }
            } else {
                if ($prefix !== null) {
                    $build->setEnvVar('PKG_CONFIG_PATH', $prefix.'/lib/pkgconfig');
                }

                return '--with-curl';
            }
```

So, the only thing I really change is set the `PKG_CONFIG_PATH` environment variables.

```php
$build->setEnvVar('PKG_CONFIG_PATH', $prefix.'/lib/pkgconfig');
```

And, why should I do this, because for none standard system library just like `homebrew` on mac, even `brewPrefixFinder` can find it. The php configure can't not find `pkgconfig` unless you set it `path` in environment. But, it will pollute the system default.

Actually, I also want the flexibility, I also plan accept `$val` in `7.4`, so the code will looks like

```php
            if ($build->compareVersion('7.4') < 0) {
                if ($val) {
                    return "--with-curl=$val";
                }

                if ($prefix !== null) {
                    return '--with-curl=' . $prefix;
                }
            } else {
                if ($val) {
                    $build->setEnvVar('PKG_CONFIG_PATH', $val.'/lib/pkgconfig');
                }
        
                if ($prefix !== null) {
                    $build->setEnvVar('PKG_CONFIG_PATH', $prefix.'/lib/pkgconfig');
                }

                return '--with-curl';
            }
```

Then, I can pass my personal build or none standard curl/libxml2/openssl to phpbrew, and build `php-7.4.x` withouts problems.

src/PhpBrew/VariantBuilder.php
markwu: Please see the big comment below.

src/PhpBrew/VariantBuilder.php
markwu: Please see the big comment below.

src/PhpBrew/VariantBuilder.php
markwu: Please see the big comment below.

markwu:
Here comes the `brew info openssl`:

```bash
$ brew info openssl
openssl@1.1: stable 1.1.1d (bottled) [keg-only]
Cryptography and SSL/TLS Toolkit
https://openssl.org/
/usr/local/Cellar/openssl@1.1/1.1.1d (7,983 files, 17.9MB)
  Poured from bottle on 2019-12-03 at 02:20:51
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/openssl@1.1.rb
==> Caveats
A CA file has been bootstrapped using certificates from the system
keychain. To add additional certificates, place .pem files in
  /usr/local/etc/openssl@1.1/certs

and run
  /usr/local/opt/openssl@1.1/bin/c_rehash

openssl@1.1 is keg-only, which means it was not symlinked into /usr/local,
because openssl/libressl is provided by macOS so don't link an incompatible version.

If you need to have openssl@1.1 first in your PATH run:
  echo 'export PATH="/usr/local/opt/openssl@1.1/bin:$PATH"' >> ~/.bash_profile

For compilers to find openssl@1.1 you may need to set:
  export LDFLAGS="-L/usr/local/opt/openssl@1.1/lib"
  export CPPFLAGS="-I/usr/local/opt/openssl@1.1/include"

For pkg-config to find openssl@1.1 you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig"
```

You can see it says:

```bash
openssl@1.1 is keg-only, which means it was not symlinked into /usr/local,
because openssl/libressl is provided by macOS so don't link an incompatible version.
```

src/PhpBrew/Build.php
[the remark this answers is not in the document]
markwu: Done!

src/PhpBrew/Build.php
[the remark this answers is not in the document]
markwu: Done!

src/PhpBrew/Build.php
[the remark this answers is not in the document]
markwu: Done!

src/PhpBrew/Command/InstallCommand.php
markwu: Dones

src/PhpBrew/VariantBuilder.php
```diff
@@ -315,34 +315,27 @@ public function __construct()
...
-            }
-            if ($bin = Utils::findBin('curl-config')) {
-                if ($prefix = exec_line("$bin --prefix")) {
-                    if (file_exists($prefix)) {
-                        return "--with-curl=$prefix";
-                    }
-                    echo "homebrew prefix '$prefix' doesn't exist. you forgot to install?\n";
+            } else {
+                if ($prefix !== null) {
+                    $build->setPkgConfigPath($prefix.'/lib/pkgconfig');
```
morozov: I'm still struggling to understand the logic being implemented. By the moment when this line of code is reached, I assume the following:

1. We're on macOS.
2. We determined the build prefix (most likely, from Homebrew by using the `BrewPrefixFinder`).
3. Besides passing the `curl` build prefix explicitly to the PHP build scripts, we're passing it the `PKG_CONFIG_PATH` based on this prefix.

The questions are:
1. What is it supposed to use this prefix for?
2. How are prefixes from multiple extensions are supposed to work together?
3. Is this fix specific to these couple of extensions or it's applicable to any Homebrew-based dependencies?
4. What is the specific change in PHP 7.4 that requires this additional logic and why is it needed only for macOS?

morozov:
> And, why should I do this, because for none standard system library just like `homebrew` on mac, even `brewPrefixFinder` can find it.

This is confusing. Isn't the sole purpose of the `BrewPrefixFinder` to find dependencies managed by Homebrew on macOS? If it doesn't do its job properly, can we fix it and enable all extensions that use it instead fixing only a couple of extensions by adding more duplicated code?

markwu:
Okay, let me explain this way.

`macOS` has it's own building system from `Xcode` and `CLT` (command line environment). It has its own building tools, libraries and packages. It official and system wide.

`homebrew` is another building system, it also has its own building tools, libraries and packages.

To build PHP outside `homebrew`  with homebrew's packages. You need to link them to system directories. For example, `/usr/local/bin`, `/usr/local/lib`, `/usr/local/include` and etc.

But, to avoid problems, `homebrew` only link missed libraries to system, and won't link same libraries to system. You have to do it manually.

Take `openssl` for example, macOS has it's own openssl:
```bash
# it is system wide
$ /usr/bin/openssl version
LibreSSL 2.6.5
```
`homebrew`  has openssl 1.1.1d
```bash
# it is only available inside homebrew building system, unless you link to system directories
$ /usr/local/opt/openssl/bin/openssl version
OpenSSL 1.1.1d  10 Sep 2019
```
Okay, if we build PHP inside homebrew, it will find `pkg-config` information from `/usr/local/opt/openssl/bin/openssl/lib/pkgconfig` automatically without any problems.

But, if we install PHP outside homebrew, like phpbrew. There is no way to find `pkg-config` information unless we set `PKG_CONFIG_PATH=/usr/local/opt/openssl/bin/openssl/lib/pkgconfig`.

So, even `BrewPrefixFinder` can find openssl, but if the `configure` doesn't understand them, there is no way to compile.

Before PHP 7.3(with), we can pass the openssl `prefix` to php configure with `--with-openssl=$(brew --prefix openssl)`, therefore there is no problem to compile it.

After PHP 7.4(with), there is no way to pass openssl prefix to php configure, because there is only `--with-openssl` available, you can't pass anything into configure. We need to pass `pkg-config`, `ldflags`, `cppflags` through environment variables.

That's why after 7.4, we have to export `PKG_CONFIG_PATH` manually.

The purpose of this PR:

1. Pass the `$val` from  phpbrew command or `$prefix` from `XXXPrefixFinder` to `--with-openssl=($val|$prefix)`  before php 7.3.

2. Pass the `$val` from  phpbrew command or `$prefix` from `XXXPrefixFinder` to `PKG_CONFIG_PATH` with `--with-openssl`  after php 7.4.

Hope this explanation is clear for you.

morozov:
Thank you for this detailed writeup. It really helps understand the changes.

> After PHP 7.4(with), there is no way to pass openssl prefix to php configure, because there is only `--with-openssl` available, you can't pass anything into configure. We need to pass `pkg-config`, `ldflags`, `cppflags` through environment variables.

For future reference, it happened in https://github.com/php/php-src/pull/3654.

src/PhpBrew/Build.php
```diff
@@ -98,6 +98,19 @@ public function setOSRelease($osRelease)
         $this->osRelease = $osRelease;
     }
 
+    public function setPkgConfigPath($path)
```
morozov: Since this method effectively collects the paths, let's redefine it as:
```php
/** @var array<string,true> */
private $pkgConfigPaths = array();

public function getPkgConfigPaths()
{
    return array_keys($this->pkgConfigPaths);
}

public function addPkgConfigPath($path)
{
    $this->pkgConfigPaths[$paths] = true;
}

```

This way, the paths will be collected and naturally duplicated, and then whatever component should use these paths, could call `$build->getPkgConfigPaths()` and pass them to the `configure` command. This way, instead of just setting the environment variable globally, it would be better manageable and testable.

src/PhpBrew/VariantBuilder.php
morozov: Should we check if this path exists before passing it to `configure`?

src/PhpBrew/VariantBuilder.php
```diff
@@ -315,34 +315,27 @@ public function __construct()
...
                 }
+
+                return '--with-curl';
```
morozov: This `return` is redundant. If it's omitted, the result will be the same.

src/PhpBrew/VariantBuilder.php
```diff
@@ -570,33 +563,26 @@ public function __construct()
...
+            } else {
+                if ($prefix !== null) {
+                    $build->setPkgConfigPath($prefix.'/lib/pkgconfig');
+                }
 
-            $possiblePrefixes = array('/usr/local/opt/openssl');
-            $foundPrefixes = array_filter($possiblePrefixes, 'file_exists');
-            if (count($foundPrefixes) > 0) {
-                return '--with-openssl=' . $foundPrefixes[0];
+                return '--with-openssl';
```
morozov: This `return` is redundant. If it's omitted, the result will be the same.

src/PhpBrew/VariantBuilder.php
```diff
@@ -570,33 +563,26 @@ public function __construct()
...
             }
 
             // This will create openssl.so file for dynamic loading.
```
morozov: This comment together with the following `echo` seems no longer valid. Could you please remove them and the `return '--with-openssl';` above?

morozov:
> Because the current extension manager install them by `phpize`, `configure`, `make`. Without the capability like `VariantBuilder` add extra settings for you...

This is a shortcoming of the current design. In fact, all the prefixes should be determined on the per-extension basis regardless whether it's `phpbrew install` or `phpbrew ext install` while variants would be combinations of extensions or their synonyms. We could approach refactoring this logic in the future.

src/PhpBrew/Build.php
```diff
@@ -98,6 +98,19 @@ public function setOSRelease($osRelease)
...
+    {
+        $value = getenv('PKG_CONFIG_PATH');
+        if ($value === "") {
+            putenv('PKG_CONFIG_PATH='.$path);
+        } else {
+            $pkgConfigPaths = explode(PATH_SEPARATOR, $value);
```
morozov: If the original `PKG_CONFIG_PATH` is empty, it will result in the array containing one empty string instead of an empty array. So the resulting `PKG_CONFIG_PATH` will be `:/new/path` instead of `/new/path`.

src/PhpBrew/Build.php
markwu: I can do this, but I need another method to put them to shell environment, for example `setPkgConfigPaths()`. Since we don't have `Environment` class, maybe we can add it to `Build` class, `Config` class, or `VariantBuilder` itself. I am not sure which one is better.

And, I can call it after we build options from `$option=$variantBuilder->build($build)`.

The pseudo code may looks like: 

```php
       // InstallCommand.php 
       // add method setPkgConfigPaths() inside Build class, call it after VariantBuilder build $option
        $variantBuilder = new VariantBuilder();
        $configureOptions = $variantBuilder->build($build);
        $build->setPkgConfigPaths());
```
or
```php
        // InstallCommand.php 
        // add method putPkgConfigPaths() inside Config class, call it after VariantBuilder build $option
        $variantBuilder = new VariantBuilder();
        $configureOptions = $variantBuilder->build($build);
        Config::setPkgConfigPaths($build->getPkgConfigPaths());
```

src/PhpBrew/VariantBuilder.php
markwu: I'll remove it.

src/PhpBrew/VariantBuilder.php
markwu: I'll remove it.

src/PhpBrew/Build.php
markwu: I will remove the empty string element first.

morozov: If you rework it as suggested in https://github.com/phpbrew/phpbrew/pull/1085#discussion_r358589687, it won't be needed.

src/PhpBrew/Build.php
markwu: Fixed!

src/PhpBrew/VariantBuilder.php
markwu: Removed!

src/PhpBrew/Build.php
markwu: Done.

markwu:
I just fixed the issues we discussed.

src/PhpBrew/Build.php
```diff
@@ -83,6 +90,35 @@ public function __construct($version, $name = null, $installPrefix = null)
...
+        $this->pkgConfigPaths[$path] = true;
+    }
+
+    public function putPkgConfigPathsEnv()
+    {
+        $paths = $this->getPkgConfigPaths();
+        $pathCount = count($paths);
+
+        if ($pathCount == 0) {
+            return false;
```
morozov: This return value isn't used anywhere, so the condition itself isn't relevant as well. Please remove the `if`/`else` and return and leave only the code under `else`.

src/PhpBrew/Build.php
```diff
@@ -83,6 +90,35 @@ public function __construct($version, $name = null, $installPrefix = null)
...
+        } else {
+            $currentPkgConfigPath = getenv('PKG_CONFIG_PATH');
+            if ($currentPkgConfigPath !== '') {
```
morozov: It can also be `false`. Please take that into account:
```sh
php -r "var_dump(getenv('BLAH'));"
bool(false)
```

src/PhpBrew/VariantBuilder.php
```diff
@@ -1180,6 +1175,10 @@ public function build(Build $build)
         // reset options
         $this->options = array();
 
+        // before return all options, we need to put the PKG_CONFIG_PATH
+        // from variants to shell environment.
+        $build->putPkgConfigPathsEnv();
```
morozov: The `VariantBuilder::build()` method is meant to build the variant and not produce any side effects. Is it possible to move this method call outside of this method somewhere in the `InstallCommand` closer to where the `ConfigureTask` is executed?

morozov:
> ```bash
> $ phpbrew --debug install --jobs "$(nproc)" 7.1.33 +default +dbs +dba +apxs2 +bz2="$(brew --prefix bzip2)" +zlib="$(brew --prefix zlib)" +ftp
> ```

BTW, from this command, it follows that PHPBrew doesn't properly discover the Homebrew-managed prefixes for `zlib` and `bz2`. Do you want to file another PR and rework these variants to use the prefix finders and employ the one for Homebrew?

markwu:
> > ```shell
> > $ phpbrew --debug install --jobs "$(nproc)" 7.1.33 +default +dbs +dba +apxs2 +bz2="$(brew --prefix bzip2)" +zlib="$(brew --prefix zlib)" +ftp
> > ```
> 
> BTW, from this command, it follows that PHPBrew doesn't properly discover the Homebrew-managed prefixes for `zlib` and `bz2`. Do you want to file another PR and rework these variants to use the prefix finders and employ the one for Homebrew?

Yes, I plan to use another PR for this, because it seems different issue.

src/PhpBrew/VariantBuilder.php
markwu: Okay, I see.

src/PhpBrew/Build.php
markwu: I'll check it.

src/PhpBrew/Build.php
```diff
@@ -103,20 +103,13 @@ public function addPkgConfigPath($path)
...
-        } else {
-            $currentPkgConfigPath = getenv('PKG_CONFIG_PATH');
-            if ($currentPkgConfigPath !== '') {
-                $currentPaths = explode(PATH_SEPARATOR, $currentPkgConfigPath);
-                $paths = array_unique(array_merge($paths, $currentPaths));
-            }
-            putenv('PKG_CONFIG_PATH='.implode(PATH_SEPARATOR, $paths));
-
-            return true;
+        if (!empty($currentPkgConfigPath)) {
```
morozov: It's better to keep it `currentPkgConfigPath !== false`. `0` is a valid path and cannot be ignored

morozov:
Please address the CI failure. You'll want to rebase your branch first because some of the code style issues are already addressed in the upstream branch. Then please squash all commits.

markwu:
@morozov I just fixed the issues you mentioned, and squashed all commits into one. Thanks for your review.

I will fire another PR for add `BrewPrefixFinder` support for `iconv` and `bz2`.

morozov:
Thank you @markwu!

morozov:
I just checked this approach on macOS and it seems not working. First I tried `phpbrew install` and then just ran a raw `./configure` command:
```
$ brew info libxml2
libxml2: stable 2.9.10 (bottled), HEAD [keg-only]
GNOME XML library
http://xmlsoft.org/
/usr/local/Cellar/libxml2/2.9.10 (280 files, 10.5MB)
  Poured from bottle on 2019-12-23 at 16:22:01
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/libxml2.rb
==> Dependencies
Required: python ✔, readline ✔
==> Options
--HEAD
	Install HEAD version
==> Caveats
libxml2 is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have libxml2 first in your PATH run:
  echo 'set -g fish_user_paths "/usr/local/opt/libxml2/bin" $fish_user_paths' >> ~/.config/fish/config.fish

For compilers to find libxml2 you may need to set:
  set -gx LDFLAGS "-L/usr/local/opt/libxml2/lib"
  set -gx CPPFLAGS "-I/usr/local/opt/libxml2/include"

==> Analytics
install: 57,321 (30 days), 143,787 (90 days), 630,154 (365 days)
install-on-request: 15,167 (30 days), 37,983 (90 days), 133,004 (365 days)
build-error: 0 (30 days)

$ env PKG_CONFIG_PATH=/usr/local/opt/libxml2/lib/pkgconfig ./configure '--with-libxml'

...

Configuring extensions
checking io.h usability... no
checking io.h presence... no
checking for io.h... no
checking for strtoll... yes
checking for atoll... yes
checking whether to build with LIBXML support... yes
checking for libxml-2.0 >= 2.7.6... no
configure: error: in `/Users/morozov/.phpbrew/build/php-7.4.1':
configure: error: The pkg-config script could not be found or is too old.  Make sure it
is in your PATH or set the PKG_CONFIG environment variable to the full
path to pkg-config.

Alternatively, you may set the environment variables LIBXML_CFLAGS
and LIBXML_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.

To get pkg-config, see <http://pkg-config.freedesktop.org/>.
See `config.log' for more details
```
I've never used PHPBrew on macOS so I may be doing something wrong. Any ideas?

morozov:
The same is if I do:
```
env LDFLAGS="-L/usr/local/opt/libxml2/lib" CPPFLAGS="-I/usr/local/opt/libxml2/include" ./configure
```

markwu:
I think there are some homebrew dependency need to installed first to run phpbrew, for example `pkg-config`. Can you check if you installed `pkg-config` through `brew list |  grep pkg-config`, it should be:

```bash
$ brew list | grep pkg-config
pkg-config
```

I build a formula before to install `phpbrew` for myself, you can install the dependency list in this script https://github.com/markwu/homebrew-php/blob/master/Formula/phpbrew.rb

```ruby
  depends_on "autoconf"
  depends_on "automake"
  depends_on "curl"
  depends_on "gd"
  depends_on "gettext"
  depends_on "gmp"
  depends_on "icu4c"
  depends_on "bison"
  depends_on "jpeg"
  depends_on "libevent"
  depends_on "libtool"
  depends_on "libxml2"
  depends_on "mcrypt"
  depends_on "mhash"
  depends_on "openssl"
  depends_on "pcre"
  depends_on "re2c"
```

Maybe we can host our own homebrew formula to install all these dependencies with `phpbrew` easier.

Mark

markwu:
you can try the following  steps to install phpbrew by homebrew:

1. brew tap markwu/php 
2. brew update
3. brew install phpbrew
4. phpbrew self-update
5. You can copy your own version phpbrew to here homebrew folder with `cp phpbrew /usr/local/opt/phpbrew/libexec/` 
6. phpbrew --debug install --jobs "$(nproc)" 7.4.0 +default +dbs +dba +apxs2 +ftp +soap +gettext +exif +iconv +bz2 +zlib

morozov:
> I think there are some homebrew dependency need to installed first to run phpbrew, for example `pkg-config`.

Exactly. It was missing on my machine. After I installed it, I was able to build PHP 7.4. Added it to the [requirements](https://github.com/phpbrew/phpbrew/wiki/Requirement) page.

morozov:
> Maybe we can host our own homebrew formula to install all these dependencies with `phpbrew` easier.

I think what we could do instead is to document all [variant dependencies](https://github.com/phpbrew/phpbrew/wiki/Variant-Dependencies) for all supported platforms and implement integration of PHPBrew with dependency managers (including Homebrew). This way, it could just run or offer to run the needed commands w/o the need to install additional software. But this is a really long-going plan.

theofidry:
Thanks a lot of the work! 🥇

PR #1118, opened 2019-12-22
Add homebrew support for bz2 and iconv

markwu:
Add homebrew dependency support for bz2 and iconv. Now, you don't need to do `+bz2=$(brew --prefix bz2)` and `+iconv=$(brew --prefix iconv)`, just `+bz2` and `+iconv` is enough.

But, this only works for `phpbrew install`, not for `phpbrew ext install`, you still need to assign prefix as before.

### DIFF
--- a/src/PhpBrew/VariantBuilder.php
+++ b/src/PhpBrew/VariantBuilder.php
@@ -844,30 +844,37 @@ class VariantBuilder
             return '--with-gettext';
         };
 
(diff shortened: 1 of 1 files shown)


morozov:
> But, this only works for `phpbrew install`, not for `phpbrew ext install`, you still need to assign prefix as before.

This will need to be taken care of separately. Thanks @markwu!

PR #1132, opened 2019-12-26
Use internal oniguruma to build mbstring/mbregex in previous PHP version before 7.4.0

markwu:
Use internal `oniguruma` to build mbstring/mbregex in previous PHP version before 7.4.0

### DIFF
--- a/src/PhpBrew/VariantBuilder.php
+++ b/src/PhpBrew/VariantBuilder.php
@@ -162,7 +162,7 @@ class VariantBuilder
         $this->variants['mbstring'] = function (ConfigureParameters $params, Build $build, $value) {
             $params = $params->withOption('--enable-mbstring');
 
-            if ($build->compareVersion('5.4') >= 0 && !$build->isDisabledVariant('mbregex')) {
+            if ($build->compareVersion('7.4') >= 0 && !$build->isDisabledVariant('mbregex')) {
                 $prefix = Utils::findPrefix(array(
                     new UserProvidedPrefix($value),
                     new IncludePrefixFinder('oniguruma.h'),


src/PhpBrew/VariantBuilder.php
```diff
@@ -162,7 +162,7 @@ public function __construct()
         $this->variants['mbstring'] = function (ConfigureParameters $params, Build $build, $value) {
             $params = $params->withOption('--enable-mbstring');
 
-            if ($build->compareVersion('5.4') >= 0 && !$build->isDisabledVariant('mbregex')) {
+            if ($build->compareVersion('7.4') >= 0 && !$build->isDisabledVariant('mbregex')) {
```
morozov: It's not a typo. See the [documentation](https://www.php.net/manual/en/mbstring.installation.php):
> As of PHP 5.4.0, if Oniguruma is already installed on the system, `--with-onig[=DIR]` can be specified to use the installed library.

markwu: I see, Thanks. I'll close this PR.

markwu:
PHP-5.6.40 and PHP-7.3.13 ships it's own oniguruma, but PHP-7.4.1 doesn't. It located under

~/.phpbrew/build/php-5.6.40/ext/mbstring/oniguruma
~/.phpbrew/build/php-7.3.13/ext/mbstring/oniguruma

I guess PHP-7.(0|1|2) ships with it's ownoniguruma, too.

But, PHP-5.6.40 can only build --enable-mbregex with it's shipped oniguruma, and php-7.3.13 can build --enable-mbregex with both shipped and external oniguruma

So, as you said, maybe we should reopen #1132 to force build previous PHP version (Before PHP-7.4.0) with it's own shipped oniguruma.

That's why homebrew's php formula has additional oniguruma dependency after PHP-7.4.0.

I think I should change the commit to `Use internal oniguruma before PHP-7.4.0` instead of `Fixed version typo`.

morozov:
Thanks, @markwu!

morozov:
Backported to `1.25` via 14d78195cc4744826dbb75c25fcced09e4f7e6e9.

Issue #1134, opened 2019-12-27
phpbrew config edit the wrong version


markwu:
Here comes my `phpbrew list` and `phpbrew system` results:

```
$ phpbrew list
  php-7.4.1
  php-7.3.13
* php-5.6.40
$ php -v
PHP 5.6.40 (cli) (built: Dec 27 2019 17:15:44)
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans
$ phpbrew system
/Users/mark/.phpbrew/php/php-7.3.13/bin/php
```

It use `PHP-5.6.40` as default php interpreter, and use `PHP-7.3.13` as system interpreter. When I use `phpbrew config` to edit my `php.ini` file. I suppose it will edit php.ini for 5.6.40, but it just edit php.ini for 7.3.13.

I also try `phpbrew list-ini`, it shows me ini files for 7.3.13, instead of 5.6.40.
```
$ phpbrew list-ini
Loaded ini files:
 - /Users/mark/.phpbrew/php/php-7.3.13/var/db/ast.ini
 - /Users/mark/.phpbrew/php/php-7.3.13/var/db/gd.ini
 - /Users/mark/.phpbrew/php/php-7.3.13/var/db/opcache.ini
 - /Users/mark/.phpbrew/php/php-7.3.13/var/db/xdebug.ini
```

But, `phpbrew  cd` works well.
```
$ phpbrew cd etc
Switching to /Users/mark/.phpbrew/php/php-5.6.40/etc, run 'cd -' to go back.
```
Is it expected behavior with the new introduced `system` command? or a bug?

My phpbrew version is `1.25.1` and os is `macOS 10.14.6`.

morozov:
Yeah, this is a bug. Besides being used as an interpreter for PHPBrew, the system interpreter shouldn't leak into any other use cases.

morozov:
It looks like in order to work properly, the functionality of `ConfigCommand` and `ListIniCommand` needs to be reimplemented in the shell. Otherwise, even if we execute them using the currently `use`d interpreted, it will stop working when we bump the runtime PHP version requirement.

Ideally, we want to identify potentially affected commands like those before reworking them.

markwu:
[...]

PR #1135, opened 2019-12-28
Fixed system interpreter leaked issue #1134.

markwu:
Use exec() to run php code with used interpreter instead of system interpreter.

### DIFF
--- a/src/PhpBrew/Command/ConfigCommand.php
+++ b/src/PhpBrew/Command/ConfigCommand.php
@@ -15,7 +15,7 @@ class ConfigCommand extends Command
 
     public function execute()
     {
-        $file = php_ini_loaded_file();
+        $file = $this->getCurrentPhpLoadedIniFile();
         if (!file_exists($file)) {
             $php = Config::getCurrentPhpName();
             $this->logger->warn("Sorry, I can't find the {$file} file for php {$php}.");
@@ -24,5 +24,21 @@ class ConfigCommand extends Command
         }
 
         Utils::editor($file);
+    }
+
+    private function getCurrentPhpLoadedIniFile()
+    {
+        $cmd  = Config::getCurrentPhpBin() . '/php';
+        $options = ' -r ';
+        $code = '"echo php_ini_loaded_file();"';
+
+        exec($cmd . $options . $code, $output, $retVal);
+        if ($retVal) {
+            return false;
+        }
+
+        $file = $output[0];
+
+        return $file;
     }
 }
--- a/src/PhpBrew/Command/ConfigCommand.php
+++ b/src/PhpBrew/Command/ConfigCommand.php
@@ -16,7 +16,9 @@ class ConfigCommand extends Command
 
     public function execute()
     {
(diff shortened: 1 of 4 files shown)


markwu:
The result:

```
$ phpbrew list
* php-7.4.1
  php-7.3.13
  php-5.6.40

$ phpbrew system
/Users/mark/.phpbrew/php/php-7.3.13/bin/php

$ phpbrew list-ini
Loaded ini files:
 - /Users/mark/.phpbrew/php/php-7.4.1/var/db/ast.ini
 - /Users/mark/.phpbrew/php/php-7.4.1/var/db/gd.ini
 - /Users/mark/.phpbrew/php/php-7.4.1/var/db/opcache.ini
 - /Users/mark/.phpbrew/php/php-7.4.1/var/db/xdebug.ini
```

markwu:
`phpbrew info` has the same problem, too.

markwu:
I try to wrap the function, so I can get the function call through used php interpreter like following. We can get exact the same result as before.

```php
UsePhpFunctionWrapper::execute('php_ini_loaded_file()', $output) ? $file = $output : $file = '';
```

```php
<?php

namespace PhpBrew;
use PhpBrew\Config;

class UsePhpFunctionWrapper
{
    public static function execute($func, &$output = null)
    {
        $retVal = false;

        $command = Config::getCurrentPhpBin() . '/php';

        $descriptorspec = array(
           0 => array("pipe", "r"),
           1 => array("pipe", "w"),
           2 => array("pipe", "w")
        );

        $process = proc_open($command, $descriptorspec, $pipes);

        if (is_resource($process)) {
            fwrite($pipes[0], '<?php echo serialize(' . $func  .'); ?>');
            fclose($pipes[0]);

            $output = stream_get_contents($pipes[1]);
            fclose($pipes[1]);

            $retVal = proc_close($process);

            if ($retVal === 0) {
                $output = unserialize($output);

                return true;
            }
        }

        return false;
    }
}
```

markwu:
`phpbrew config`, `phpbrew list-ini` and `phpbrew info` fixed.

morozov:
This looks even more complicated than I thought it would. We can do it much simpler:
1. `phpbrew config` could be implemented as:
   ```sh
   # Bash
   $ command $EDITOR "$(php -r "echo php_ini_loaded_file();")"

   # Fish
   $ command $EDITOR (php -r "echo php_ini_loaded_file();")
   ```
2. `phpbrew info` currently excutes a series of subsequent calls to the underlying PHP and glues the output together. Instead, we could do something like:
   ```
   # phpbrew info-script generates some script
   $ phpbrew info-script
   <?php
   
   echo 'Version' . PHP_EOL;
   echo 'PHP-' . phpversion() . PHP_EOL;
   // and so on

   # its output gets piped to the currently used PHP
   phpbrew info-script | php
   ```

markwu:
Yes, you are right, it's much simpler. I just tried to keep everything as before.

But, I am not so familiar with the shell scripts. I will leave the implement for you, and see If I can catch how you do it.

morozov:
Closing in favor of #1136.

PR #1158, opened 2020-02-11
Support fish v2

gucchisk:
## Purpose
Support fish v2 for CentOS and so on.
(In CentOS7, version of fish which is installed by yum is 2.3.1)

If it is not necessary to support fish v2, I'll close this PR...

## Changes
- Use `and` & `or` instead of `&&` & `||`
- Halt the use of `-a` option in `set`
- Use `eval` instead of `command`, because it is not supported by fish2 to execute variable.
- Add validation of arguments at `each`

### DIFF
--- a/shell/phpbrew.fish
+++ b/shell/phpbrew.fish
@@ -259,7 +259,9 @@ function phpbrew
             return 0
 
         case each
-            __phpbrew_each $argv[2..-1]
+	    if [ (count $argv) -gt 1 ]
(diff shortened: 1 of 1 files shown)


morozov:
I’d rather not support two major versions of a shell. Otherwise, it would require testing all changes in two different shells w/o any automation.

gucchisk:
Oh, I see.
Thank you for you response.

PR #1161, opened 2020-02-14
Remove references to AppStore in README

anubisthejackle:
Because of the removal of the AppStore, `phpbrew app` no longer exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,22 +425,6 @@ phpbrew fpm config
 >
 >     php-fpm --php-ini {php config file} --fpm-config {fpm config file}
 
-## Installing Extra Apps
-
-phpbrew provides app command to fetch some php apps.
-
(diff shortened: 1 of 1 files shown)


morozov:
Thank you @anubisthejackle!

tpraxl:
I just spent some time to find out why the `phpbrew app` command is not working as documented on the website. Now I know. It has been removed. The documentation needs to be removed from the website as well: https://github.com/phpbrew/phpbrew/tree/gh-pages

abenevaut:
Does it means we have to install composer app outside phpbrew or is there another way to install it inside phpbrew ?

morozov:
PHPBrew no longer adds its `.phpbrew/bin` to `$PATH`, so there's no longer “inside phpbrew”.

PR #1167, opened 2020-03-11
Remove references to phpbrew app from gh-pages

anubisthejackle:
Removes reference to `phpbrew app` from gh-pages


morozov:
Thank you, @anubisthejackle!

PR #1171, opened 2020-03-31
Fix intl compilation for php7.4.x in macOS

markwu:
We should use withPkgConfigPath for icu4c in php7.4.x or compilation will failed.


morozov:
Thanks, @markwu!

PR #1190, opened 2020-08-02
PhpStreamDownloader: throw exceptions on failure

smola:
Execution continued unnecessarily when there is a download failure, leading to long and confusing traces.


morozov (approved):

morozov:
Thanks, @smola.

Issue #1165, opened 2020-03-09
Different configurations for the same version of php


halaei:
### 1. Different cli and fpm configurations:
I want to disable max_memory, mad max_execution_time limits for cli but not keep the settings for fpm. I guess currently there is no way to do so because there is only one php.ini. Also #820 is a related issue, requiring the ability of disabling an extension for cli.

### 2. Multiple instances of the same version:
Although not 100% storage efficient, to solve the mentioned problem above, I can also install multiple instances of the same version, one with cli, the other with fpm, but I don't see how this is possible using phpbrew. Having multiple instances can also help us benefit from php 7.4 preloading, to preload different classes for different projects.

 Thanks for your help

PR #1195, opened 2020-09-21
PHP will be built using different configs for different SAPIs

### LINKED ISSUE #820: Disabling an extension specifically for php-cli
When installing extensions with `phpbrew ext install`, `phpbrew` builds and installs the requested extension and loads it by adding an ini-file to var/db. This makes it easy to install, disable or enable extensions. However, since the _var/db_ path is probably hard-configured during the build process of php, there's no way to disable an extension for the cli version of php.

In this case, I'm specifically referring to the xdebug extension, which causes a lot of overhead for things like composer. Is there a way for disabling an extension specifically for php-cli, while keeping it enabled for the apache/(f)cgi/fpm version of php? Of course, preferrably without messing with the workings of `phpbrew`, for instance by adding the `zend_extension=xdebug.so` line to the php.ini instead of the ini-file in var/db and keeping it out of php-cli.ini.

biozshock:
Build php with different php.ini for each SAPI: apache2, fpm, cli. Extensions draw their config from the different directories too.
Fixes #820 #1165

### DIFF
--- a/completion/bash/_phpbrew
+++ b/completion/bash/_phpbrew
@@ -2037,8 +2037,8 @@ local comp_prefix=___phpbrew
         subcommands=()
 subcommand_alias=()
 subcommand_signs=()
-options=()
-options_require_value=()
+options=(["-s"]="1" ["--sapi"]="1" )
+options_require_value=(["-s"]="1" ["--sapi"]="1" )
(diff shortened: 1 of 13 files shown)


c9s (approved):
LGTM

PR #1250, opened 2021-09-28
Fix phpbrew bash prompt for php 8

jrmgx:
This is a basic fix.
Note: I did not find the related code for other shell prompt (they may even not have a similar option)
if it exists it could be done in a separate PR

### DIFF
--- a/shell/bashrc
+++ b/shell/bashrc
@@ -535,7 +535,7 @@ function __phpbrew_purge_build ()
 
 function phpbrew_current_php_version() {
   if type "php" > /dev/null; then
-    local version=$(php -v | grep -E "PHP [57]" | sed 's/.*PHP \([^-]*\).*/\1/' | cut -c 1-6)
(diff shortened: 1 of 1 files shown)


jhdxr (approved):
LGTM

jhdxr:
Thanks. The fish shell script has similar functions (https://github.com/phpbrew/phpbrew/blob/bd6d47cdfc9c64d7384be057ecaba8ea6e2ba317/shell/phpbrew.fish#L487) but no worry, I'll update it.

jhdxr:
hmm, I think travis-ci.org is dead. @c9s Can you help to configure it optional?

peter279k:
> hmm, I think travis-ci.org is dead. @c9s Can you help to configure it optional?

As far as I know, Travis CI change travis-ci.org to travis-ci.com for about whiles. And it's not friendly for open source projects since then.

IMHO, I think it's time to switch to the GitHub Actions now :).

c9s:
let me remove the travis ci require

c9s:
thank you all!

PR #1293, opened 2022-12-15
Update GitHub Actions workflow

theofidry:
- Update some GitHub Actions pre-defined steps
- Use [ramsey/composer-install](https://github.com/ramsey/composer-install) for installing Composer dependencies which offers built-in caching

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,12 +8,12 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest]
-        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1']
+        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3-dev']
     name: PHP ${{ matrix.php-versions }} Test on ${{ matrix.operating-system }}
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v3
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
@@ -23,10 +23,10 @@ jobs:
           coverage: none
 
       - name: Install dependencies
-        run: composer install
+        uses: ramsey/composer-install@v2
 
       - name: Update PHPUnit
         run: composer update phpunit/phpunit --with-dependencies
 
       - name: Run test suite
-        run: vendor/bin/phpunit --exclude-group mayignore --coverage-clover build/logs/clover.xml
+        run: vendor/bin/phpunit --exclude-group mayignore
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest]
-        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3-dev']
+        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1']
     name: PHP ${{ matrix.php-versions }} Test on ${{ matrix.operating-system }}
 
     steps:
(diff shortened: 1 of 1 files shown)


peter279k:
It seems that the CI checking is failed during the `PHP 8.2` tests running.

Is the internal `PHP 8.2` error?

.github/workflows/build.yml
```diff
@@ -8,12 +8,12 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest]
-        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1']
+        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3-dev']
```
peter279k, line 11: `8.3` is the unstable version and I think we don't test the development version.

peter279k: > It seems that the CI checking is failed during the `PHP 8.2` tests running.
> 
> Is the internal `PHP 8.2` error?

I also get the failed test in my local machine:

```sh
$ php -v
PHP 8.2.0 (cli) (built: Dec 10 2022 10:52:23) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.0, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.0, Copyright (c), by Zend Technologies
$ vendor/bin/phpunit --exclude-group mayignore
vendor/bin/phpunit --exclude-group mayignore
PHPUnit 8.5.31 by Sebastian Bergmann and contributors.

................SSSSSSS..SSS......................W............  63 / 306 ( 20%)
......................Segmentation fault (core dumped)
```

peter279k: After digging the issue, it seems that the enabling VCR with [php-vcr](https://github.com/php-vcr/php-vcr) is problematic.

peter279k: After removing related VCR enabling/disabling codes, the test suites are worked successfully.

peter279k: I also describe the https://github.com/php-vcr/php-vcr/issues/373 on this repository.

peter279k: At this moment, I think we don't support PHP `8.2` version.

Please remove the PHP `8.2` test during GitHub action.

.github/workflows/build.yml
```diff
@@ -23,10 +23,10 @@ jobs:
...
 
       - name: Install dependencies
-        run: composer install
+        uses: ramsey/composer-install@v2
 
       - name: Update PHPUnit
         run: composer update phpunit/phpunit --with-dependencies
 
       - name: Run test suite
-        run: vendor/bin/phpunit --exclude-group mayignore --coverage-clover build/logs/clover.xml
```
peter279k, line 32: Please keep the code coverage report option.

It will be used when I integrate the `php-coveralls` step during the GitHub action running in the upcoming PR.

.github/workflows/build.yml
theofidry: @peter279k I we are mixing up different issues here.

>At this moment, I think we don't support the PHP 8.2 version.

IMO this is not because a version is run in the CI that it means it's supported. I think it is beneficial to add it as if you want to work on PHP 8.2 support, then having it as part of the CI is a nice to have.

I also think I PHPBrew doesn't support PHP 8.2 & 8.3, then it should convey it from the `composer.json`. The current constraint `^8.0` indicates otherwise. Would you be fine with fixing this?

Last but not least: I see that there is no required check in the pipeline. I recommend to enable them: Settings -> Branches -> Add a rule or edit the rule for the `master` branch and:
- tick  "Require status checks to pass before merging"
- look for status checks to check

This way you have a distinction between a build that is required, and one that is purely informative (e.g. 8.2 or 8.3 compatibility)

peter279k: Thanks for your reply. I think we can keep the PR until these versions are supported and compatible.

The PHP version definition will be fixed in another PR until the failed dependency is resolved or another solution is figured out.

theofidry: My goal was really only to do a simple update not to spark debate to be honest. I didn't think including non supported upcoming PHP versions was a contentious point. So I'll remove this for now and we can address this in a different PR.

c9s (approved):

PR #1322, opened 2023-02-04
ci: Add a release workflow

theofidry:
Add a release workflow which builds the PHAR and, when triggered by a release event, will sign and publish the PHAR.

:warning: This requires to:

- Replace `<TODO:privatekeyID>` by the actual GPG key ID (you can see it when running `$ gpg -K`) and `<TODO:email>` by the email address associated with the GPG key used.
- You then need to add two environment variables `GPG_KEY_X` and `GPG_KEY_X_PASSPHRASE` to the GitHub Actions.

Note that the PHAR asc will be `phpbrew.phar.asc` instead of `phpbrew.asc`. I do not thing it is too important, but this follows the convention followed by [PHIVE](phar.io) which is widely adopted (at least by PHIVE itself, PHPUnit, Infection, PHP-Scoper & Box to name few).


c9s (approved):

c9s:
@peter279k could you help the GPG key setup? you can setup your own. thanks!

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+            # Smoke test
+            -   name: Ensure the PHAR works
+                run: bin/phpbrew.phar --version
+
+            -   name: Import GPG key
+                if: github.event_name == 'release'
+                uses: crazy-max/ghaction-import-gpg@v5
+                with:
+                    gpg_private_key: ${{ secrets.GPG_KEY_<TODO:privatekeyID> }}
```
peter279k, line 44: - I've setup the `GPG_KEY`. It can use the `secrets.GPG_KEY` now.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
...
+                    passphrase: ${{ secrets.GPG_KEY_<TODO:privatekeyID>_PASSPHRASE }}
```
peter279k, line 45: I don't have the GPG key passphrase and I think the setting can keep blank.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
...
+        steps:
+            -   uses: actions/download-artifact@v3
+                with:
+                    name: phpbrew-phar
+                    path: .
+
+            -   name: Upload phpbrew.phar
+                uses: softprops/action-gh-release@v1
+                with:
+                    token: ${{ secrets.GITHUB_TOKEN }}
```
peter279k, line 81: @c9s, could you generate the secret GITHUB_TOKEN in this repository?

theofidry: To double check but I think this one is _already_ provided by GitHub (when not in a PR)

theofidry:
Thanks! Will come back to this ASAP

.github/workflows/release.yaml
peter279k: You're right.

The GITHUB_TOKEN is a special access token that you can use to authenticate on behalf of GitHub Actions.

GitHub automatically creates a GITHUB_TOKEN secret for you to use in your workflow, and you can use it to authenticate in a workflow run.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+            -   name: Sign the PHAR
+                if: github.event_name == 'release'
+                run: |
+                    gpg --local-user <TODO:email> \
+                        --batch \
+                        --yes \
+                        --passphrase="${{ secrets.GPG_KEY_<TODO:privatekeyID>_PASSPHRASE }}" \
+                        --detach-sign \
+                        --output bin/phpbrew.phar.asc \
```
peter279k, line 55: To be consistency, I think it should be the `bin/phpbrew.asc`.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
...
+                if: github.event_name == 'release'
+                uses: crazy-max/ghaction-import-gpg@v5
+                with:
+                    gpg_private_key: ${{ secrets.GPG_KEY_<TODO:privatekeyID> }}
+                    passphrase: ${{ secrets.GPG_KEY_<TODO:privatekeyID>_PASSPHRASE }}
+
+            -   name: Sign the PHAR
+                if: github.event_name == 'release'
+                run: |
+                    gpg --local-user <TODO:email> \
```
peter279k, line 50: The e-mail cab be `peter279k@gmail.com`.

.github/workflows/release.yaml
theofidry: Note that tools like PHIVE enforces the convention `path/to/your.phar` & `path/to/your.phar.asc`; but that can be visited separately

theofidry:
@peter279k although I'm not 100% sure it works. Unfortunately it has to be tested with a "dummy release". Either on a fork repository or this repository with empty patch releases

peter279k:
@theofidry, I see that the `Release/Build PHAR` action is failed and the error log is as as follows:

```
box compile
make: box: No such file or directory
make: *** [Makefile:8: phpbrew] Error 127
```

Could you check that? Thanks.

theofidry:
@peter279k sorry for the very late reply.

Indeed the issue is that we need Box, so we need a way to get it installed. I think there is two solutions:

- Either rely on https://phar.io/, in which case one can install it as he/she wishes locally and in the CI it can be added via the `tools` section. Then we only need the phive.xml which will install Box.
- Add Box to the repository

I went for the latter as it is simple, the barrier of entry is lower. But this adds a fat binary to the repository which is not ideal tbh.

Do you have a preference about the solution to adopt?

peter279k:
@theofidry, we need the `Box` before building the Phar and there're two ways to install the box tool:

- Using the phive to install the box.
- Using the `composer require` command to install the box when running the released step CI.

In my opinion, I think these two ways are proper and I also prefer the second way because it can use the command with installing specific version.

I think we need to install/maintain the phive tool if using the first way.

theofidry:
@peter279k then this PR should be ready :)

theofidry:
@peter279k it should be ready now. Don't forget to do a "squash & merge" rather than a simple merge

PR #1323, opened 2023-02-04
ci: Rename the `build` workflow to `tests` and add a check status job

theofidry:
This PR:

- Renames the `build` workflow to `tests` which better describes what the workflow actually does.
- Add a meta job `Tests Status`. This job is executed once the whole test matrix is finished and succeeds only and only if all the jobs created by the matrix are successful. This allows to add this job, with a static name, to the protection rules of the branch. As a result, one can easily update the matrix and its label without having to require to constantly update the branch protection rules.


theofidry:
ping @peter279k, I would really like to start adding more tests & stuff to the CI soonish so that we can also more easily prevent issues #1325

theofidry:
ping @peter279k? Or do you dislike those changes?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,61 @@
...
+                uses: ramsey/composer-install@v2
+
+            -   name: Update PHPUnit
+                run: composer update phpunit/phpunit --with-dependencies
+
+            -   name: Run test suite
+                run: vendor/bin/phpunit --exclude-group mayignore --coverage-clover build/logs/clover.xml
+
+
+    # This is a "trick", a meta task which does not change, and we can use in
```
peter279k, line 45: I don't ensure the meta task is necessary. And I think it should print some status messages before running the `exit` command.

theofidry: if you are curious, this looks like [this](https://github.com/theofidry/console/actions/runs/6292164585).

You can see that "Test 8.1; Symfony 5.4" which change based on the matrix values (of course they could be fixed but then you loose information).

If you want to add a protection rule to your branch that all the tests must pass, you will need to list them all. So you're rule will look like this:

Screenshot 2023-09-30 at 22 19 48

This is not ideal because it means whenever you change the matrix parameters, your rule configuration becomes outdated. Instead the above allows you to do something like this:

Screenshot 2023-09-30 at 22 20 43

Then it doesn't matter what value you have in your matrix, as long as they all pass, the check status will pass. If any fail, it won't, and the rule will always be up to date.

theofidry: @peter279k not convinced? Should we close this then?

peter279k: Go ahead with this and I think there's no problem now.

peter279k (approved):

PR #1353, opened 2023-11-24
Fix: Prevent extra-options from being overwritten

PPP01:
The extraoptions get partially overwritten:
--with-config-file-path
--with-config-file-scan-dir

This means that the additional inis are no longer loaded from the correct directory. If you have set up many php versions, then the config-file-scan-dir per version is not practical.

In version 1.27 the path was not yet overwritten and worked because it was appended at the end of the Makefile. Since v.1.28 the path is fixed in the code and there is no possibility to overwrite it.

With the patch, the extraoptions are split if necessary and then set correctly via withOption(option,value). Before the build, the system checks whether the two options are set.

### DIFF
--- a/src/PhpBrew/Build.php
+++ b/src/PhpBrew/Build.php
@@ -8,12 +8,12 @@ use PhpBrew\BuildSettings\BuildSettings;
  * A build object contains version information,
  * variant configuration,
  * paths and an build identifier (BuildId).
- *
  * @method array getEnabledVariants()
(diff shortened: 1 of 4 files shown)


PPP01:
I'll try. But I have my difficulties with the unit tests because my dev-system does not have all needed dependencies

fuerst:
This PR is still necessary to get extraoptions ending up correctly in the created PHP binary.

Issue #1352, opened 2023-11-06
"patch for fix readline detection" is broken


Gaibhne:
The following error occurs (and is being silently ignored!) in my install and can be found in many logs and issues by people, both here, on Stack Overflow and many other places where people post logs about other problems. I believe most people don't even notice the error because it is ignored, which may or may not be intentional.

I don't know the relevance of the error, nor whether it even applies (I did not attempt to install a PHP in the given range), but if it isn't an actual error, the output with the red error message should probably be changed. As it is, I have an installation that had an error during its compilation, so I don't know whether it's all working fine or whether I will run into strange, unexplained issues in the future while using it.

### Output

```
Checking patch for fix readline detection on PHP 5.3 to PHP 7.1
---> Applying patch...
patching file ext/readline/config.m4
Hunk #1 FAILED at 60.
Hunk #2 FAILED at 121.
2 out of 2 hunks FAILED -- saving rejects to file ext/readline/config.m4.rej
```

### Expected Result

```
I expected no error message
```

### Command

Paste your phpbrew install command here (be sure to install with `--debug` flag, e.g. `phpbrew --debug install`)

    phpbrew --debug install -j $(nproc) 8.1.13 +pdo +mysql +iconv +mbstring +json +filter +openssl +tidy +zip +intl +xml +ctype

### Build Log

    The build log appears to include lots of personal information such as my real name and my companies name. Is there any way to sanitize it short of going through all of it manually myself with search & replace ? Probably not relevant anyways

### Platform

OS:

    Ubuntus LTS 22 on WSL 2

Running PHP: I don't know what to put here

Installing PHP: I don't know what to put here

peter279k:
- "Running PHP" is about which PHP version you run the PHPBrew.
- "Installing PHP" is about which PHP version you want to install. According to the `Command`, the installing PHP is `8.1.13`.

Gaibhne:
Thanks, updated

peter279k:
Closed via #1357.

PR #1357, opened 2023-12-07
fix: Resolve readline patch attempting to apply to php >7.1 due to OR instead of AND

driskell:
This resolves a patch failure (that interestingly gets ignored) when installing PHP > 7.1
There's an OR so effectively match() always returns true - since it will always either be higher or equal to 5.3 or lower or equal to 7.1

### DIFF
--- a/src/PhpBrew/Patches/ReadlinePatch.php
+++ b/src/PhpBrew/Patches/ReadlinePatch.php
@@ -19,8 +19,8 @@ class ReadlinePatch extends Patch
 
     public function match(Buildable $build, Logger $logger)
     {
-        return version_compare($build->getVersion(), '5.4', '>=')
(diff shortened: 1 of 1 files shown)


peter279k:
It seems that it's related to the https://github.com/phpbrew/phpbrew/issues/1352.

driskell:
@peter279k Yes that's related - this would fix that issue

peter279k (approved):

peter279k:
@c9s, could you help me to review this PR? Thanks.

frostfire64:
Is there any possibility for this issue to be resolved, it seems that phpbrew is a nice tool that I would like to use on WSL ubuntu but it crashes for this very reason.

peter279k:
Sorry for the delay. I've merged this PR now.

Issue #1140, opened 2019-12-30
__phpbrew_php_exec:15: command too long: phpbrew ()


deep12650:
phpbrew list is working without source ~/.phpbrew/bashrc but phpbrew switch php-7.2.26 gives error "Invalid argument php-7.2.26". Once I do  source ~/.phpbrew/bashrc  it starts giving an error "__phpbrew_php_exec:15: command too long: phpbrew ()"

### Output

```
deepsachan@Deeps-MBP ~ % phpbrew list
* (system)
  php-7.2.26
deepsachan@Deeps-MBP ~ % phpbrew switch php-7.2.26
Invalid argument php-7.2.26
deepsachan@Deeps-MBP ~ % source ~/.zshrc
deepsachan@Deeps-MBP ~ % phpbrew switch php-7.2.26
Invalid argument php-7.2.26
deepsachan@Deeps-MBP ~ % source ~/.phpbrew/bashrc
usage: mkdir [-pv] [-m mode] directory ...
deepsachan@Deeps-MBP ~ % phpbrew list
[...]

PR #1363, opened 2024-02-13
Use `whence -p` instead of `which` for Zsh

jarodhayes:
This is to address a regression to an issue outlined in #1140, and addressed with PR #1142. The fix to the issue was removed in commit 729420dadd2d034ef5b5cdcc87c0018d78209b19.

To my understanding, `which` incorrectly resolves to the phpbrew shell function instead of the binary when using Zsh. Instead, `whence -p` should be used to resolve the binary.

### DIFF
--- a/shell/bashrc
+++ b/shell/bashrc
@@ -22,7 +22,13 @@ MIN_PHP_VERSION_ID=70200
 # Returns the absolute path corresponding to the command excluding the alias
 function __phpbrew_which()
 {
-    command which "$1"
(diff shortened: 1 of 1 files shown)


jarodhayes:
Hello,

I did some additional troubleshooting and found that the issue was primarily that PHPBrew's `bashrc` was for some reason being sourced through zsh's emulation mode, specifically ksh emulation. While the pull request did fix the issue in that use case, it's probably best to just source the file directly as designed.

PR #1374, opened 2024-06-23
ci: Leverage shivammathur/setup-php to install Box in the CI

theofidry:


.github/workflows/release.yaml
```diff
@@ -20,17 +20,14 @@ jobs:
             -   name: Setup PHP
                 uses: shivammathur/setup-php@v2
                 with:
-                    php-version: '8.1'
+                    php-version: '8.2'
```
peter279k, line 23: Is there any reason about changing `8.1` to the `8.2` version?

PR #1375, opened 2024-06-23
ci: Add an AutoReview workflow

theofidry:
Introduces a workflow for the purpose of validating and reviewing the code. This where I would typically include:

- Static analysis
- Various validations, e.g. that the `composer.json` is valid (the only check currently implemented in this PR)
- Code linting
- Any unit tests that aim at checking the codebase architecture/structure/conventions, i.e. anything that is not testing the source code behaviour.

### DIFF
--- a/.github/workflows/auto-review.yaml
+++ b/.github/workflows/auto-review.yaml
@@ -1,0 +1,42 @@
+name: AutoReview
+
+on:
+    push:
+        branches: [ main, master ]
+    pull_request: ~
+
+jobs:
+    composer-validate:
+        runs-on: ubuntu-latest
+        name: Composer Validate
+        steps:
+            -   name: Checkout
+                uses: actions/checkout@v4
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    php-version: '8.2'
+                    tools: composer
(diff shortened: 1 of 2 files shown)


.github/workflows/auto-review.yaml
```diff
@@ -0,0 +1,42 @@
+name: AutoReview
+
+on:
+    push:
+        branches: [ main, master ]
```
peter279k, line 5: Could you explain why the `main` branch is added in the `branches` setting?

peter279k:
Thanks for your PR. I think the `composer.lock` will not be changed in the PR. Could you check that?

theofidry:
@peter279k clarified and extracted the composer update in #1379

peter279k:
Due to the PR about updating the `composer.lock` file is extracted from this, I suggest the `composer.lock`  should be reverted in this PR.

.github/workflows/auto-review.yaml
theofidry: Currently, there is no reason.

I would like however to rename `master` to `main` at some point, allowing the two will allow to do that any time without any issue. Note that the other added workflow have the same too

theofidry:
done

PR #1376, opened 2024-06-23
build: Reconfigure dependabot

theofidry:
- Configure it to be executed weekly only
- Check GitHub Actions too

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,16 +1,19 @@
 version: 2
 updates:
-- package-ecosystem: composer
-  directory: "/"
-  schedule:
-    interval: daily
-    time: "10:00"
-  open-pull-requests-limit: 10
-  reviewers:
-  - c9s
-  - morozov
-  assignees:
-  - morozov
-  labels:
-  - Dependencies
-  versioning-strategy: increase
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,16 +1,19 @@
...
-  open-pull-requests-limit: 10
-  reviewers:
-  - c9s
-  - morozov
-  assignees:
-  - morozov
-  labels:
-  - Dependencies
-  versioning-strategy: increase
+    -   package-ecosystem: "composer"
```
peter279k, line 3: Some settings are removed. Could you add them back?

Such as `open-pull-requests-limit`, `reviewers` , `assignees` and `labels`.

theofidry: > open-pull-requests-limit

I don't think it's necessary: there won't be that many and in my experience, with a similar config, dependabot will close and re-open MRs so there will not be many in parallel.

>reviewers, assignees

I don't think it's up to date, and is there really a point? However has access can check it out anytime instead.

>labels

The PRs already get `build(deps): ` as a name prefix, if you still want the label we can add it, I just saw no value in it.

PR #1380, opened 2024-06-24
build(tools): Introduce Rector

theofidry:
This PR introduces [Rector](https://getrector.com).

It is a tool I've been using moderately for the past few years and IMO can help our either punctually for some migration tasks or for some menial tasks.

Here I configured it with:

- Import symbols when it can (i.e. make use of `use` statements instead of using a FQCN).
- Remove unneeded imports (IMO this is more a task of a CS-Fixer but it does not hurt to have it here either, and there is no CS fixer at the moment).
- Ensure the code is PHP 7.2 compatible. This way if you accidentally add a non PHP 7.2 compatible code, this will be flagged or fixed.

Locally you can use:

```
# Execute rector lint
make rector_lint

# Execute rector "fix"
make rector
```

### DIFF
--- a/.github/workflows/auto-review.yaml
+++ b/.github/workflows/auto-review.yaml
@@ -1,0 +1,69 @@
+name: AutoReview
+
+on:
+    push:
+        branches: [ main, master ]
+    pull_request: ~
+
+jobs:
+    composer-validate:
+        runs-on: ubuntu-latest
+        name: Composer Validate
+        steps:
+            -   name: Checkout
+                uses: actions/checkout@v4
+
+            -   name: Setup PHP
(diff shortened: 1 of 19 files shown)


.github/workflows/auto-review.yaml
```diff
@@ -0,0 +1,69 @@
...
+                with:
+                    php-version: '8.2'
+                    tools: composer
+                    coverage: none
+
+            -   run: make composer_validate
+
+    rector_lint:
+        runs-on: ubuntu-latest
+        name: Composer Validate
```
theofidry, line 27:
```suggestion
        name: Rector Lint
```

PR #1381, opened 2024-06-24
build: Improve the Makefile

theofidry:


.github/workflows/release.yaml
```diff
@@ -31,9 +31,12 @@ jobs:
             -   name: Install Box
                 run: composer global require --dev humbug/box
 
+            -   name: Ensure the dependencies are up to date
+                run: make vendor
```
theofidry, line 35: this is purely to avoid noise. It can be that the `vendor` install does not have the right timestamp in which case calling `make build` will trigger another `composer install`. It is not a big deal, but it is noisy, so extracting this in a dedicate step helps to not pollute the important step bellow.

Makefile
```diff
@@ -1,25 +1,37 @@
-TARGET        = phpbrew
-SIGNATURE     = $(TARGET).asc
+# See https://tech.davis-hansson.com/p/make/
+MAKEFLAGS += --warn-undefined-variables
```
theofidry, line 2: See the link for the effect of those flags

Makefile
```diff
@@ -1,25 +1,37 @@
...
+MAKEFLAGS += --no-builtin-rules
+
+PHPBREW_PHAR        = phpbrew
```
theofidry: renamed `TARGET` to `PHPBREW_PHAR` for readability

Makefile
```diff
@@ -1,25 +1,37 @@
...
+SIGNATURE     = $(PHPBREW_PHAR).asc
 CP            = cp
 INSTALL_PATH  = /usr/local/bin
 TEST          = phpunit
 
-$(TARGET): vendor $(shell find bin/ shell/ src/ -type f) box.json.dist .git/HEAD
-	box compile
-	touch -c $@
+PHAR_SRC_FILES := $(shell find bin/ shell/ src/ -type f)
```
theofidry, line 10: Extracted `$(shell find bin/ shell/ src/ -type f)` into a dedicated variable

Makefile
```diff
@@ -1,25 +1,37 @@
...
+
+
+.DEFAULT_GOAL := help
```
theofidry, line 13: Introduce a help command. I don't think it is wise to display all the commands there, but it will be good to document the most important ones. For now I've only documented the building PHAR one.

Makefile
```diff
@@ -1,25 +1,37 @@
...
+.PHONY: help
+help:
+	@printf "\033[33mUsage:\033[0m\n  make TARGET\n\n\033[32m#\n# Commands\n#---------------------------------------------------------------------------\033[0m\n\n"
+	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//' | awk 'BEGIN {FS = ":"}; {printf "\033[33m%s:\033[0m%s\n", $$1, $$2}'
 
-vendor: composer.lock
-	composer install
-	touch $@
 
-.PHONY: sign
```
theofidry, line 15: the `sign` and `SIGNATURE` because now the PHAR is signed directly in the CI so there is no need for it in the Makefile anymore.

Makefile
```diff
@@ -1,25 +1,37 @@
...
-sign: $(SIGNATURE)
+.PHONY: build
+build:	## Builds PHPBrew PHAR
+build:
```
theofidry, line 24: pretty much the previous `make phpbrew`, with the difference that it will always compile the PHAR regardless of whether the PHAR was already built and up to date.

You also have `_build`, which I call "shadow command" and mimicks `build`, but there it does not care about the targets so it will nto rebuild the PHAR if not necessary.

I like to have this kind of command + shadow command to have a good UX whilst being able to keep shortucts of one needs it.

For example, in my projects I often have a `make serve` which will start the webserver, but will also boot up docker if not already booted up, install deps if not installed, migrate the DB if it is not up to date etc. It's a lot of stuff, and sometimes I just want to boot up the webserver without caring about all of that, and that's where I use a "shadow command", i.e. I would have `make _serve` that would just do that.

Note that here in this case, `_build` is effectively just an alias for `phpbrew` so might no be super useful. We can do without really

Makefile
```diff
@@ -29,4 +41,24 @@ test:
...
+
+$(PHPBREW_PHAR): vendor \
+		$(PHAR_SRC_FILES) \
+		box.json.dist \
+		.git/HEAD
+	box compile
+	touch -c $@
+
+PHONY: vendor_install
+vendor_install:
```
theofidry, line 53: I added some extra targets to make sure the `vendor` is up to date.

In general one can use `composer install` as usual, but if you want to force it to make sure the Make targets are up to date then you can call `make vendor_install`, and it will update all the targets.

theofidry (commented):
This PR proposes various changes, feel free to point out the elements you don't like. I can also break it up more if you prefer, I didn't want to do that initially because as a reviewer this would also make it a bit hard to see where this is going and the feedback loop is quite slow due to our different availabilities and timezone differences. So having a slightly bigger PR makes it a bit easier to collect all the necessary feedback IMO

PR #1399, opened 2025-06-09
feat: Allow quick version switching with a non-patch version prefix eg. 8.1 or 8.2

driskell:
This allows the `use` command to switch without a minor version.

So you can just simply write `use 8.1`


theofidry (approved):
I assume it would require to updated the fishrc too

